PR #25, opened 2021-09-09
Add `environment` property to context

fenix20113:
Add  the `environment` property to context to be able to: 
1. Use the [Constrain on a specific environment](https://docs.getunleash.io/advanced/strategy_constraints#constrain-on-a-specific-environment) 
2. Create custom strategies based on the environment.

### DIFF
--- a/src/Configuration/Context.php
+++ b/src/Configuration/Context.php
@@ -4,7 +4,9 @@ namespace Unleash\Client\Configuration;
 
 /**
  * @method string|null getHostname()
+ * @method string|null getEnvironment()
  * @method Context     setHostname(string|null $hostname)
+ * @method Context     setEnvironment(string|null $environment)
  */
 interface Context
 {
--- a/src/Configuration/UnleashContext.php
+++ b/src/Configuration/UnleashContext.php
@@ -17,6 +17,7 @@ final class UnleashContext implements Context
         private ?string $sessionId = null,
         private array $customContext = [],
         ?string $hostname = null,
+        private ?string $environment = null,
     ) {
         $this->setHostname($hostname);
     }
@@ -24,6 +25,11 @@ final class UnleashContext implements Context
     public function getCurrentUserId(): ?string
     {
         return $this->currentUserId;
+    }
+
+    public function getEnvironment(): ?string
+    {
+        return $this->environment;
     }
 
     public function getIpAddress(): ?string
@@ -89,6 +95,13 @@ final class UnleashContext implements Context
         return $this;
     }
 
+    public function setEnvironment(?string $environment): self
+    {
+        $this->environment = $environment;
+
+        return $this;
+    }
+
     public function getHostname(): ?string
     {
         return $this->findContextValue('hostname') ?? (gethostname() ?: null);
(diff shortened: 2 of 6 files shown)


src/Configuration/Context.php
```diff
@@ -14,6 +14,8 @@ public function getIpAddress(): ?string;
 
     public function getSessionId(): ?string;
 
+    public function getEnvironment(): ?string;
```
RikudouSage: This and `setEnvironment()` are a BC break which I would like to avoid, add them as class phpdoc methods for now.

src/Configuration/UnleashContext.php
```diff
@@ -15,6 +15,7 @@ public function __construct(
         private ?string $currentUserId = null,
         private ?string $ipAddress = null,
         private ?string $sessionId = null,
+        private ?string $environment = null,
```
RikudouSage: Another BC break, new parameters can only be added at the end of the method

src/DefaultUnleash.php
```diff
@@ -32,7 +33,8 @@ public function __construct(
 
     public function isEnabled(string $featureName, ?Context $context = null, bool $default = false): bool
     {
-        $context ??= $this->configuration->getContextProvider()->getContext();
+        $defaultContext = $this->configuration->getContextProvider()->getContext();
+        $context = $context ? self::mergeContext($defaultContext, $context) : $defaultContext;
```
RikudouSage: Why is this necessary? I can't really see the point of merging with default context instead of either using custom or the default.

src/Configuration/Context.php
fenix20113: I'll do it

src/Configuration/UnleashContext.php
fenix20113: I'll do it

src/DefaultUnleash.php
fenix20113: Seems you're right. I just realized that we do not need it. But, the initial idea was taken from the node client. https://github.com/Unleash/unleash-client-node/blob/master/src/unleash.ts#L213
```
isEnabled(name: string, context: Context = {}, fallback?: FallbackFunction | boolean): boolean {
    const enhancedContext = { ...this.staticContext, ...context };
...
```

ivarconr: The idea is that there exist a few "static" context fields (`appName` and `environment`) that does not change for a given application per request. It is thus convenient to not have to pass these for every call just because the `userId` changes. 

All other official SDKs currently follow this pattern. One could of course argue that this logic can be handled outside the SDK.

RikudouSage: I see, makes sense, but it's anyway out of scope of this PR IMHO.

fenix20113: Agree. I removed this change from the PR and pushed fixes for tests.

RikudouSage:
The current error is because the environment is not set in `\Unleash\Client\Tests\ClientSpecificationTest::createContext()`. That's why I would advise to do it similarly to how the hostname is handled in context:

```php

    public function getHostname(): ?string
    {
        return $this->findContextValue(ContextField::HOSTNAME) ?? (gethostname() ?: null);
    }

    public function setHostname(?string $hostname): self
    {
        if ($hostname === null) {
            $this->removeCustomProperty(ContextField::HOSTNAME);
        } else {
            $this->setCustomProperty(ContextField::HOSTNAME, $hostname);
        }

        return $this;
    }
```

And in constructor you will just call `$this->setEnvironment($environment)` while removing the `private` modifier from the constructor parameter. But it's up to you, I'll accept both solutions.

fenix20113:
Thanks for the explanation. I updated `\Unleash\Client\Tests\ClientSpecificationTest::createContext()`

RikudouSage:
Thank you for your contribution!

PR #27, opened 2021-09-14
Use php-http/discovery to find PSR-17/18 implementations

Xenonym:
This project relies on PSR-17 request factories and PSR-18 HTTP clients to perform HTTP requests. It will load Guzzle and Symfony implementations if they exist, otherwise the user is required to provide their own.

Let's use [`php-http/discovery`](https://github.com/php-http/discovery) to load more PSR-17/18 implementations automatically. This will reduce the need for users to configure their own PSR-17/18 implementations manually.

### DIFF
--- a/src/Helper/DefaultImplementationLocator.php
+++ b/src/Helper/DefaultImplementationLocator.php
@@ -3,8 +3,9 @@
 namespace Unleash\Client\Helper;
 
 use Cache\Adapter\Filesystem\FilesystemCachePool;
-use GuzzleHttp\Client;
-use GuzzleHttp\Psr7\HttpFactory;
+use Http\Discovery\Exception\NotFoundException;
+use Http\Discovery\Psr17FactoryDiscovery;
+use Http\Discovery\Psr18ClientDiscovery;
 use League\Flysystem\Adapter\Local;
 use League\Flysystem\Filesystem;
 use LogicException;
@@ -13,7 +14,6 @@ use Psr\Http\Message\RequestFactoryInterface;
 use Psr\SimpleCache\CacheInterface;
 use Symfony\Component\Cache\Adapter\FilesystemAdapter;
 use Symfony\Component\Cache\Psr16Cache;
-use Symfony\Component\HttpClient\Psr18Client;
 
 /**
  * @internal
@@ -42,14 +42,6 @@ final class DefaultImplementationLocator
      * @var array<string,array<string,array>>
      */
     private array $defaultImplementations = [
-        'client' => [
-            Client::class => [],
-            Psr18Client::class => [],
-        ],
-        'factory' => [
-            HttpFactory::class => [],
-            Psr18Client::class => [],
-        ],
         'cache' => [
             FilesystemCachePool::class => [
                 Filesystem::class => [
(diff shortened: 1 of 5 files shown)


RikudouSage:
Thanks, I was planning to look into it myself! Just please try to maintain the 100% code coverage. Also the logic around supported packages for http services is now obsolete and should be removed since any implementation should be supported.

Xenonym:
@RikudouSage thanks for reviewing! I thought I had updated the `testWIthoutDefaultClients` to maintain code coverage, but turns out I misunderstood how the `Psr18ClientDiscovery` and `Psr17FactoryDiscovery` classes work, so the code path for having a PSR-18 client but _not_ a PSR-17 request factory is never called.

I am not sure how this could be tested though, due to how `php-http/discovery` works. The discovery for a PSR-18 client starts with checking if the `Symfony\Component\HttpClient\Psr18Client` and `Psr\Http\Message\RequestFactoryInterface` classes exist: 

https://github.com/php-http/discovery/blob/778f722e29250c1fac0bbdef2c122fa5d038c9eb/src/Strategy/CommonClassesStrategy.php#L98-L101

Since both `symfony/http-client` and `nyholm/psr7` are required in this package, the class check for `Psr\Http\Message\RequestFactoryInterface` will always succeed, _even if_ you try to remove the PSR-17 discovery strategy like I attempted in the `testWIthoutDefaultClients` test. Doing so results in an unexpected `DiscoveryFailedException` as the discovery attempts to construct a `Psr18Client` as it believes that the PSR-17 request factory interface is available (and thus a PSR-17 implementation must be available).  In an actual project that only has `symfony/http-client` and no PSR-17 implementations, the class check will not actually succeed, so the above scenario will not happen.

I think you could test this specific scenario of having a PSR-18 implementation without a PSR-17 implementation by mocking `DefaultImplementationLocator` with an implementation that could specifically avoid returning a PSR-17 request factory only. However, I wonder if it's worth the effort - I believe `symfony/http-client` is the _only_ well-known PSR-18 client that requires you to bring your own PSR-17 implementation. We could also ignore coverage for this section altogether - what do you think?

RikudouSage:
@Xenonym It seems that the code coverage for that might be unnecessary, would you mind putting the `@code-coverage-ignore-start` etc. tags in the code? I'd like to have every line of this package covered or marked explicitly as ignored (in this case some short description in comment would also be nice, possibly with link to this issue).

Xenonym:
@RikudouSage marked the areas as requested! I think you will need to approve the workflows to trigger again 😅

src/Helper/DefaultImplementationLocator.php
```diff
@@ -68,40 +52,35 @@ final class DefaultImplementationLocator
...
+            return Psr17FactoryDiscovery::findRequestFactory();
+        } catch (NotFoundException $e) {
+            /**
+             * This will only be thrown if a HTTP client was found, but a request factory is not.
+             * Due to how php-http/discovery works, this scenario is unlikely to happen.
+             * See linked comment for more info.
+             *
+             * https://github.com/Unleash/unleash-client-php/pull/27#issuecomment-920764416
+             */
+            // @codeCoverageIgnoreStart
```
RikudouSage: The exception catching should be marked as uncovered as well, otherwise it's reported as missing coverage.

```suggestion
            return Psr17FactoryDiscovery::findRequestFactory();
            // @codeCoverageIgnoreStart
        } catch (NotFoundException $e) {
            /**
             * This will only be thrown if a HTTP client was found, but a request factory is not.
             * Due to how php-http/discovery works, this scenario is unlikely to happen.
             * See linked comment for more info.
             *
             * https://github.com/Unleash/unleash-client-php/pull/27#issuecomment-920764416
             */
```

thomasheartman:
[Rename GitHub default branches from master to main](https://trello.com/c/Ey6olr8s/669-rename-github-default-branches-from-master-to-main)

PR #44, opened 2022-02-04
Feat: Bootstrapping

### LINKED ISSUE #43: feat: Add support for bootstrapping feature toggles
Improve resilience by implementing the Bootstrap feature of Unleash. Feature write-up:
https://www.getunleash.io/blog/bootstrap-unleash-java-sdk-to-improve-the-resilience

Reference implementation in Java SDK:
https://github.com/Unleash/unleash-client-java/pull/131

We need to bootstrap to be adapted to what makes sense in PHP. It should at least allow:
- manual passing of "json string of the full feature toggle configuration", 
- loading from file 
- loading from URL. 

If the user want to bootstrap from a URL we should probably try to load the bootstrap content first, then the actual Unleash API, as it assumed to be faster than the Unleash API. I am also unsure how bootstrap and caching in the PHP should play together.

RikudouSage:
closes #43 

Implemented according to specifications - if the request to unleash api fails it falls back to a bootstrap provider (if present).

Also added a configuration parameter to disable fetching the api at all in which case these things are disabled:

- fetching features
- client registration
- sending metrics

If fetching is disabled and no bootstrap is provided, an exception is thrown.


src/Repository/DefaultUnleashRepository.php
```diff
@@ -51,22 +53,38 @@ public function findFeature(string $featureName): ?Feature
...
-                $request = $request->withHeader($name, $value);
+                try {
+                    $response = $this->httpClient->sendRequest($request);
+                    if ($response->getStatusCode() === 200) {
+                        $data = $response->getBody()->getContents();
+                    }
+                } catch (Exception $exception) {
+                    // empty catch, $data does not exist and will be handled below
+                }
+                $data ??= $this->getBootstrappedResponse();
```
ivarconr, line 79: Nice, I have not seen that operator before! 

Would it make sense to validate if it is anything more than an empty string? 
What if it is an empty list of feature toggles?

RikudouSage: Empty string couldn't pass here, an exception would be thrown somewhere in the providers. Empty feature toggle list is a valid response, no? For example, if your unleash server goes down you want to play it safe and don't let anyone see a list of any flag guarded features.

I can add some validator that it at least is in valid format to fail early, but I think we shouldn't check for more than the data having a "features" key and it being an array.

ivarconr: > but I think we shouldn't check for more than the data having a "features" key and it being an array.

Agreed!

ivarconr (approved):

chriswk (approved):
Great job.

src/Repository/DefaultUnleashRepository.php
RikudouSage: @ivarconr I added the check in the latest commit.

RikudouSage:
@ivarconr @chriswk Just added a commit that checks for body validity and throws exception if it isn't valid, just let me know if it's ok and I can merge.

ivarconr (approved):

ivarconr:
> @ivarconr @chriswk Just added a commit that checks for body validity and throws exception if it isn't valid, just let me know if it's ok and I can merge.

make sense.

PR #77, opened 2022-04-22
Feat: Add event system

RikudouSage:
# Description

Adds event system to the SDK, prerequisite for impression data

## Type of change

# How Has This Been Tested?

- [x] Unit tests
- [x] Integration tests
- [x] Manual tests without event dispatcher installed

# Checklist:


thomasheartman:
[GitHub PR (PHP client): Add event system](https://trello.com/c/n0S4q5sg/1098-github-pr-php-client-add-event-system)

doc/events.md
```diff
@@ -0,0 +1,308 @@
+# Event system in Unleash PHP SDK
+
+This SDK supports events using [`symfony/event-dispatcher`](https://packagist.org/packages/symfony/event-dispatcher).
+
+## Installation
+
+The event dispatcher is not a mandatory component of this SDK, so you need to install it by running:
```
ivarconr, line 7: I like that it is optional.

doc/events.md
```diff
@@ -0,0 +1,308 @@
...
+class MyEventSubscriber implements EventSubscriberInterface
+{
+    public static function getSubscribedEvents(): array
+    {
+        return [
+            UnleashEvents::FEATURE_TOGGLE_DISABLED => 'onFeatureDisabled',
+            UnleashEvents::FEATURE_TOGGLE_MISSING_STRATEGY_HANDLER => 'onNoStrategyHandler',
+            UnleashEvents::FEATURE_TOGGLE_NOT_FOUND => 'onFeatureNotFound',
+            UnleashEvents::FEATURE_VARIANT_BEFORE_FALLBACK_RETURNED => 'onVariantNotFound',
+        ];
```
ivarconr, line 34: These events are probably something we should define across SDKs.

ivarconr: (I will take a note of that, these are fine for now)

src/Event/UnleashEvents.php
```diff
@@ -0,0 +1,40 @@
...
+namespace Unleash\Client\Event;
+
+final class UnleashEvents
+{
+    /**
+     * When a feature toggle is not found this event gets triggered.
+     *
+     * You can set your own feature using ->setFeature() on the event object
+     * or change the default response using ->setEnabled().
+     *
```
ivarconr: You are taking a slightly different approach here where events is also a feedback mechanism allowing the user act on them. Other SDKs using events is only using it to inform the user about what is happening, and use other means for handling edge cases, like the "fallback function". 

To some extend this provide a lot of control and flexibility for the SDK user. At the same time it does increase the complexity. 

I would like to also get input from @sighphyre and @FredrikOseberg and get their inputs as well.

RikudouSage: True, the Symfony event system is kind of a mix between events and middlewares which is kinda powerful. I can change it to plain "dumb" events if needed. I thought that it could be used instead of the fallback function as it provides the same mechanism though by different means.

sighphyre: This is my gut feel as well. This seems to do two things, one is to bring the PHP SDK to feature parity with the Node SDKs events, the other is to add a very cool new feedback eventing system. I want to suggest we break this into two separate PRs if that's possible? One for dumb events and an optional second layered on top of that to include the reactive events.  

If the feedback eventing is currently just an alternative way to implement an existing feature I think we should be a bit careful including for two reasons: 1) it's hard to take nice things away from people and once it's in people will use it 2) it's very likey that users will want us to implement this in the other SDKs and a lot of the other SDKs don't yet have eventing. I do think this is a super cool approach though and I'd love to see if we can discuss it a bit more at some point and see if we can either use this in interesting ways or what it would take to commit to this in all the SDKs and if we all feel these would be a useful things to do. @RikudouSage What do you think?

RikudouSage: If that's the consensus I'll rewrite it to "dumb" events. These reactive events are a standard in Symfony event dispatcher, that's why I did it the way I did it.

FredrikOseberg: I think this is really cool. It's giving the user some really powerful options to interact with the SDK. I do think feature parity first is a good idea. But this is definitely a discussion we should have.

ivarconr: Lets focus on simple "dumb" events first. We can always add smarter reactive events later, but this would be something I would prefer to standardize across all SDKs a bit (and requires a bit of research on what we actually can do), it can be quite hard to maintain if the different SDKs take different approaches.

RikudouSage: @ivarconr @FredrikOseberg @sighphyre It's done, please re-review.

sighphyre: @RikudouSage Thanks, yeah, looking at the diff, this is a lot closer to what the node sdk does. I like it! I would love to see `changed` event similar to the node sdk, since a lot of consumers find that one useful, but this sets a great framework for adding that later in a separate PR. I'm happy!

sighphyre (approved):
Awesome, thanks for this!

FredrikOseberg (approved):
LGTM

ivarconr (approved):

PR #90, opened 2022-05-19
Feat: Add impression data support

RikudouSage:
# Description

Adds support for impression data.

## Type of change

# How Has This Been Tested?

- [x] Integration tests

# Checklist:

### DIFF
--- a/src/DefaultUnleash.php
+++ b/src/DefaultUnleash.php
@@ -7,10 +7,13 @@ use Unleash\Client\Configuration\Context;
 use Unleash\Client\Configuration\UnleashConfiguration;
 use Unleash\Client\DTO\Strategy;
 use Unleash\Client\DTO\Variant;
+use Unleash\Client\Enum\ImpressionDataEventType;
 use Unleash\Client\Event\FeatureToggleDisabledEvent;
 use Unleash\Client\Event\FeatureToggleMissingStrategyHandlerEvent;
 use Unleash\Client\Event\FeatureToggleNotFoundEvent;
+use Unleash\Client\Event\ImpressionDataEvent;
 use Unleash\Client\Event\UnleashEvents;
+use Unleash\Client\Helper\Uuid;
 use Unleash\Client\Metrics\MetricsHandler;
 use Unleash\Client\Repository\UnleashRepository;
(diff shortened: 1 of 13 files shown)


src/DefaultUnleash.php
```diff
@@ -49,6 +52,18 @@ public function isEnabled(string $featureName, ?Context $context = null, bool $d
             return $default;
         }
 
+        if (method_exists($feature, 'hasImpressionData') && $feature->hasImpressionData()) {
+            $event = new ImpressionDataEvent(
+                ImpressionDataEventType::IS_ENABLED,
+                Uuid::v4(),
+                clone $this->configuration,
+                clone $context,
+                clone $feature,
```
RikudouSage, line 61: It makes sense to me to send a copy and not the actual object so that it cannot be changed from within the event. In php `clone` performs a shallow copy. From [php documentation](https://www.php.net/manual/en/language.oop5.cloning.php):

> When an object is cloned, PHP will perform a shallow copy of all of the object's properties. Any properties that are references to other variables will remain references.

src/DefaultUnleash.php
```diff
@@ -113,6 +128,18 @@ public function getVariant(string $featureName, ?Context $context = null, ?Varia
...
             $this->metricsHandler->handleMetrics($feature, true, $variant);
+
+            if (method_exists($feature, 'hasImpressionData') && $feature->hasImpressionData()) {
+                $event = new ImpressionDataEvent(
+                    ImpressionDataEventType::GET_VARIANT,
+                    Uuid::v4(),
+                    clone $this->configuration,
+                    clone $context,
+                    clone $feature,
+                    clone $variant,
```
RikudouSage, line 139: Same as before, cloning the objects instead of providing the actual instances.

src/DefaultUnleash.php
sighphyre: Definitely agree, a shallow copy is probably okay here

src/DefaultUnleash.php
sighphyre: Definitely agree

sighphyre (approved):
LGTM

PR #92, opened 2022-05-30
docs: update return type documentation

rlcurrall:
# Description

This addresses deprecation warnings from Symfony regarding return types.

~~To ensure that this remains compatible with previous versions of PHP, a doc block type annotation is used for `getMetadata` instead of using type hinting with the native `mixed` pseudo type introduced in PHP 8.0~~
PHP 8.1 is the currently supported version of PHP so it is fine to use the `mixed` pseudo type.

## Type of change

Misc: Type annotation and documentation update.

# How Has This Been Tested?

As this does not modify the functionality of the code, no additional tests have been added.

# Checklist:

### DIFF
--- a/src/Helper/StringStream.php
+++ b/src/Helper/StringStream.php
@@ -48,7 +48,7 @@ final class StringStream implements StreamInterface
         return $content;
     }
 
-    public function close()
+    public function close(): void
     {
         if ($this->stream === null) {
             throw new StreamException('The stream is detached');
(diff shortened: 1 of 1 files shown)


rlcurrall:
I just checked the `composer.json` and noticed PHP 8.1 is the currently supported version, so I can add the type hint for `mixed` instead of using the doc block.

RikudouSage (approved):
Thank you!

PR #97, opened 2022-06-20
Feat: Add stale cache

RikudouSage:
# Description

Adds configuration for stale data cache triggered in case of http communication failure.

Ref https://github.com/Unleash/unleash-client-symfony/issues/23 (this is part one, the other part must be done in Symfony bundle after this gets merged).

## Type of change

# How Has This Been Tested?

- [x] Unit tests

# Checklist:

### DIFF
--- a/src/Configuration/UnleashConfiguration.php
+++ b/src/Configuration/UnleashConfiguration.php
@@ -41,6 +41,7 @@ final class UnleashConfiguration
         private bool $fetchingEnabled = true,
         // todo remove nullability in next major version
         private ?EventDispatcher $eventDispatcher = null,
+        private int $staleTtl = 30 * 60,
     ) {
         $this->contextProvider ??= new DefaultUnleashContextProvider();
         $this->eventDispatcher ??= new EventDispatcher(null);
@@ -259,6 +260,18 @@ final class UnleashConfiguration
     {
         $eventDispatcher ??= new EventDispatcher(null);
         $this->eventDispatcher = $eventDispatcher;
+
+        return $this;
+    }
+
+    public function getStaleTtl(): int
+    {
+        return $this->staleTtl;
+    }
+
+    public function setStaleTtl(int $staleTtl): self
+    {
+        $this->staleTtl = $staleTtl;
 
         return $this;
     }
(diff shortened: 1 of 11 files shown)


src/Configuration/UnleashConfiguration.php
```diff
@@ -41,6 +41,7 @@ public function __construct(
         private bool $fetchingEnabled = true,
         // todo remove nullability in next major version
         private ?EventDispatcher $eventDispatcher = null,
+        private int $staleTtl = 30 * 60,
```
RikudouSage, line 44: To anyone reviewing this, what would you expect as the default stale ttl?

src/Event/UnleashEvents.php
```diff
@@ -24,4 +24,11 @@ final class UnleashEvents
      * @Event("Unleash\Client\Event\FeatureToggleMissingStrategyHandlerEvent")
      */
     public const FEATURE_TOGGLE_MISSING_STRATEGY_HANDLER = 'unleash.event.toggle.missing_strategy_handler';
+
+    /**
+     * Triggered when fetching features from server fails.
+     *
+     * @Event("Unleash\Client\Event\FetchingDataFailedEvent")
+     */
+    public const FETCHING_DATA_FAILED = 'unleash.event.server.fetching_failed';
```
RikudouSage, line 33: Is this ok for an event name?

src/Configuration/UnleashConfiguration.php
ivarconr: this is seconds?

RikudouSage: Yep, seconds.

src/Event/UnleashEvents.php
ivarconr: looks good!

ivarconr: yes

src/Configuration/UnleashConfiguration.php
ivarconr: I think it makes sense. Wdyt @sighphyre ?

sighphyre: @RikudouSage Honestly I'm not sure I would have a number in mind, only that there was one and that if it was important to know what the number was, that it was somewhere that I would see in the docs/readme. I think what you have here makes sense to me!

sighphyre (commented):
So my PHP isn't strong but looking at the logic here this looks good to me!

Does it make sense to also include the cache times here: https://docs.getunleash.io/sdks/php_sdk ?

RikudouSage:
@sighphyre Not sure, I don't think the equivalent is listed for any other sdk?

sighphyre:
@RikudouSage So, we don't really have the equivalent for the other SDKs, from my understanding this change is particular to PHP. I thought about this a bit more though, and I think what you've done in the README is actually great and it doesn't need to be replicated in the docs

sighphyre (approved):
LGTM

PR #101, opened 2022-06-27
Enhancement: Make all events extend one parent

### LINKED ISSUE #96: Chore: Make all events extend AbstractEvent
Currently every event class file needs to check if the Symfony event dispatcher is installed and load event stub if it isn't. This should be done in one parent abstract class (introduced in #97).

RikudouSage:
# Description

Makes all parents extend one parent (AbstractEvent)

Fixes #96 

## Type of change

- [x] Enhancement

# Checklist:

### DIFF
--- a/src/Event/FeatureToggleNotFoundEvent.php
+++ b/src/Event/FeatureToggleNotFoundEvent.php
@@ -2,12 +2,7 @@
 
 namespace Unleash\Client\Event;
 
-use Symfony\Contracts\EventDispatcher\Event;
 use Unleash\Client\Configuration\Context;
-
-if (!class_exists(Event::class)) {
-    require __DIR__ . '/../../stubs/event-dispatcher/Event.php';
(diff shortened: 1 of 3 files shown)


src/Event/FeatureToggleNotFoundEvent.php
```diff
@@ -9,7 +9,7 @@
     require __DIR__ . '/../../stubs/event-dispatcher/Event.php';
```
sighphyre: Can this not also be removed here?

RikudouSage: Thank you, don't know how I missed it, it's removed now.

sighphyre: No worries! This is why we have PRs!

sighphyre (approved):
Looks good to me!

PR #104, opened 2022-07-07
Feat: Global segments

### LINKED ISSUE #103: Feature: Global Segments

### Describe the new feature

Just a feature that needs to be implemented - Global Segments. Unleash v4.13 supports enhanced responses for global segments, it would be great if this SDK can make use of this.

### Background

Segments are effectively a way for Unleash users to define a list of constraints in such a way that makes them reusable across toggles without manually copying the constraint across to another toggle. Segments have two modes of operation, from the SDK's perspective, the inline mode will have no impact, segments will be remapped on the server side into constraints on the toggle information, no changes need to be . The second mode, global segments, requires that the SDK both opt in and handle the response differently. The handling should effectively result in unpacking the segments referenced in the feature strategies into a set of constraints. The changes required are described below.

### Solution suggestions

**Control Header**

The SDK needs to pass up a `Unleash-Client-Spec` header with a semver value greater or equal to `4.2.0`  (i.e. be greater or equal to the version of the [unleash client spec tests](https://github.com/Unleash/client-specification) where global segments are described) when hitting the get toggles endpoint on the Unleash server. This will enable the Unleash server to respond with the enhanced format

Example of the difference between enhanced and standard format:

**Standard Format (default)**

```
{
   "version": 2,
   "features": [  
       {
           "strategies": [
               {
                   "name": "flexibleRollout",
                   "constraints": [
                       {
                           "values": [
                               "31"
                           ],
                           "inverted": false,
                           "operator": "IN",
                           "contextName": "appName",
                           "caseInsensitive": false
                       }
                   ],          
                   "parameters": {
                       "groupId": "Test1",
                       "rollout": "100",
                       "stickiness": "default"
                   }
               }
           ],
           "name": "Test1"
       },
       {
           "strategies": [
               {
                   "name": "flexibleRollout",
                   "constraints": [
                       {
                           "values": [
                               "31"
                           ],
                           "inverted": false,
                           "operator": "IN",
                           "contextName": "appName",
                           "caseInsensitive": false
                       }
                   ],          
                   "parameters": {
                       "groupId": "Test2",
                       "rollout": "100",
                       "stickiness": "default"
                   }
               }
           ],
           "name": "Test2"
       }    
   ],
   "query": {
       "environment": "default"
   }
}

```

***Enhanced Format (requires opt in)***
 
```
{
   "version": 2,
   "features": [   
       {
           "strategies": [
               {
                   "name": "flexibleRollout",
                   "constraints": [],
                   "parameters": {
                       "groupId": "Test1",
                       "rollout": "100",
                       "stickiness": "default"
                   },
                   "segments": [
                       1
                   ]
               }
           ],
           "name": "Test1"
       },
       {
           "strategies": [
               {
                   "name": "flexibleRollout",
                   "constraints": [],
                   "parameters": {
                       "groupId": "Test2",
                       "rollout": "100",
                       "stickiness": "default"
                   },
                   "segments": [
                       1
                   ]
               }
           ],
           "name": "Test2"
       }     
   ],
   "query": {
       "environment": "default"
   },
   "segments": [
       {
           "id": 1,           
           "constraints": [
               {
                   "values": [
                       "31"
                   ],
                   "inverted": false,
                   "operator": "IN",
                   "contextName": "appName",
                   "caseInsensitive": false
               }
           ]           
       }
   ]
}
```
The relevant changes between the two formats are that in the enhanced format the segments are defined once as a global list and referenced within the strategy on the toggle by its ID. What's important to note is that the two above packets should be 
handled identically, they reference the same toggle state.

**Considerations**

- Global segments are intended to handle large datasets, how large has not been formally specified yet but expectations are around 1 000 to 10 000 segments across 1 000 toggles. As a result, time and space complexity of the implementations needs to be considered. 
- In the case of global segments, if the mapping from segment id to segment is incomplete i.e. a segment id is referenced in a toggle strategy that doesn’t map back to the global segments list, then the toggle should be evaluated to false. This is enforced through one of the client specification tests in v4.2.0 of the client spec
- A reference implementation is provided in node JS: https://github.com/Unleash/unleash-client-node/pull/329 (note that this doesn't include the header, that can be seen here: https://github.com/Unleash/unleash-client-node/pull/335)

RikudouSage:
# Description

Global segments

Fixes #103 

## Type of change

# How Has This Been Tested?

- [x] Spec Tests

# Checklist:

### DIFF
--- a/src/Strategy/AbstractStrategyHandler.php
+++ b/src/Strategy/AbstractStrategyHandler.php
@@ -24,7 +24,16 @@ abstract class AbstractStrategyHandler implements StrategyHandler
 
     protected function validateConstraints(Strategy $strategy, Context $context): bool
     {
+        if (method_exists($strategy, 'hasNonexistentSegments') && $strategy->hasNonexistentSegments()) {
+            return false;
+        }
+
         $constraints = $strategy->getConstraints();
+        $segments = method_exists($strategy, 'getSegments') ? $strategy->getSegments() : [];
+        foreach ($segments as $segment) {
+            $constraints = [...$constraints, ...$segment->getConstraints()];
+        }
+
         foreach ($constraints as $constraint) {
             if (!$this->getValidator()->validateConstraint($constraint, $context)) {
                 return false;
--- a/src/Strategy/AbstractStrategyHandler.php
+++ b/src/Strategy/AbstractStrategyHandler.php
@@ -3,6 +3,7 @@
 namespace Unleash\Client\Strategy;
 
 use Unleash\Client\Configuration\Context;
+use Unleash\Client\DTO\Constraint;
 use Unleash\Client\DTO\Strategy;
 use Unleash\Client\Helper\ConstraintValidatorTrait;
 
@@ -28,18 +29,28 @@ abstract class AbstractStrategyHandler implements StrategyHandler
             return false;
         }
 
-        $constraints = $strategy->getConstraints();
-        $segments = method_exists($strategy, 'getSegments') ? $strategy->getSegments() : [];
-        foreach ($segments as $segment) {
-            $constraints = [...$constraints, ...$segment->getConstraints()];
-        }
+        $validator = $this->getValidator();
+        $constraints = $this->getConstraintsForStrategy($strategy);
 
         foreach ($constraints as $constraint) {
-            if (!$this->getValidator()->validateConstraint($constraint, $context)) {
+            if (!$validator->validateConstraint($constraint, $context)) {
                 return false;
             }
         }
 
         return true;
+    }
+
+    /**
+     * @return iterable<Constraint>
+     */
+    private function getConstraintsForStrategy(Strategy $strategy): iterable
+    {
+        yield from $strategy->getConstraints();
+
+        $segments = method_exists($strategy, 'getSegments') ? $strategy->getSegments() : [];
+        foreach ($segments as $segment) {
+            yield from $segment->getConstraints();
+        }
     }
 }
--- a/tests/CoverageOnlyTest.php
+++ b/tests/CoverageOnlyTest.php
@@ -1,0 +1,19 @@
+<?php
+
+namespace Unleash\Client\Tests;
+
+use PHPUnit\Framework\TestCase;
+use Unleash\Client\DTO\DefaultSegment;
+
+/**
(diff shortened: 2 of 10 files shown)


src/Strategy/AbstractStrategyHandler.php
```diff
@@ -24,7 +24,16 @@ protected function findParameter(string $parameter, Strategy $strategy): ?string
...
     protected function validateConstraints(Strategy $strategy, Context $context): bool
     {
+        if (method_exists($strategy, 'hasNonexistentSegments') && $strategy->hasNonexistentSegments()) {
+            return false;
+        }
+
         $constraints = $strategy->getConstraints();
+        $segments = method_exists($strategy, 'getSegments') ? $strategy->getSegments() : [];
+        foreach ($segments as $segment) {
+            $constraints = [...$constraints, ...$segment->getConstraints()];
```
sighphyre: This looks to my very uninformed PHP eyes that this is unpacking each segment into the array in a loop. Is this an efficient way to do this? I'm mostly asking because we have folks that are wanting to use this feature to create very large segments and reuse them a lot

tests/CoverageOnlyTest.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+namespace Unleash\Client\Tests;
+
+use PHPUnit\Framework\TestCase;
+use Unleash\Client\DTO\DefaultSegment;
+
+/**
+ * This class is only for triggering code that doesn't really make sense to test and is here to achieve 100% code coverage.
+ * The reason is to catch potential problems during transpilation to lower versions of php.
```
sighphyre, line 10: Nice, that's pretty clever!

src/Strategy/AbstractStrategyHandler.php
RikudouSage: You're right, that could be a little inefficient, don't know what I was thinking. I redid it to use yield and also got rid of one null check that wasn't necessary in each iteration.

sighphyre: Hahaha, this is why we do code review, I do things like this all the time :)

Thank you! That looks great!

sighphyre (approved):
This looks great I see all the spec test are passing, the code looks really nice, I'm happy to merge if you're happy here

RikudouSage:
Yep, good to go, thanks!

PR #111, opened 2022-08-11
Chore: Trigger uncovered code

### LINKED ISSUE #98: Chore: Trigger uncovered code
Currently simple DTOs etc. have code coverage ignored because there's nothing to test really.

The ignore statements should be removed where possible and the code should at least be triggered (without any testing) to confirm that there's no error preventing it from working (which could happen during transpilation for older versions of php).

RikudouSage:
# Description

Triggers all uncovered code that can be triggered to catch errors in transpilation.

Fixes #98 

## Type of change

-  [x] Improvement

# Checklist:


sighphyre (approved):
LGTM

PR #120, opened 2022-10-12
Allow psr/simple-cache 3.x

### LINKED ISSUE #119: Feat: Allow installation with psr/simple-cache 3.x

### Describe the new feature

I would like this library to support `psr/simple-cache` 3.x

I'm using a third-party cache adapter that requires `psr/simple-cache: ^3.0` which is preventing me from installing this library.

I could fork the cache adapter and manually make it support 1.x, or downgrade to an older version, but neither approach seems ideal.  And those workarounds wouldn't help other users who also want to use the latest PSR interfaces.

This library could easily support this new version with no `src` code changes needed - simply widening the constraints in `composer.json` would be sufficient:

```diff
-       "psr/simple-cache": "^1.0",
-       "psr/simple-cache-implementation": "^1.0",
+       "psr/simple-cache": "^1.0 | ^2.0 | ^3.0",
+       "psr/simple-cache-implementation": "^1.0 | ^2.0 | ^3.0",
```

Unfortunately, the test suite will never run on these higher versions because the `cache/filesystem-adapter` dev dependency only supports 1.x:

image

This could be worked around by making this an optional dependency (where we'd remove it from `composer.json` but conditionally `composer require` it in the CI pipeline matrix).  I'll open a PR shortly that uses this approach to better explain how it would work.

colinodell:
# Description

Adds support for using this library with `psr/simple-cache` 2.x and 3.x without breaking backward-compatibility.

Closes #119

## Type of change

# How Has This Been Tested?

- [x] Unit tests
- [x] Spec Tests

# Checklist:

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -49,6 +49,8 @@ jobs:
     strategy:
       matrix:
         version: ['8.1']
+        optional-deps: ['', 'cache/filesystem-adapter:^1.1']
+
     steps:
       - name: Setup PHP
         uses: shivammathur/setup-php@v2
@@ -58,7 +60,11 @@ jobs:
         uses: actions/checkout@v2
         with:
           submodules: true
-      - name: Install Dependencies
+      - name: Install Dependencies (without optional)
+        if: ${{ matrix.optional-deps == '' }}
         run: composer install
+      - name: Install Dependencies (with optional)
+        if: ${{ matrix.optional-deps != '' }}
+        run: composer require ${{ matrix.optional-deps }}
       - name: Run tests
         run: composer phpunit
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -1,4 +1,13 @@
 parameters:
   ignoreErrors:
-    - '#@throws with type .+ is not subtype of Throwable#'
+    -
+      message: '#@throws with type .+SimpleCache.+ is not subtype of Throwable#'
+      path: src/Repository/DefaultUnleashRepository.php
+      reportUnmatched: false
+    -
+      messages:
+        - '#Class .+Filesystem.* not found#'
+        - '#Class .+Flysystem.* not found#'
+      path: src/Helper/DefaultImplementationLocator.php
+      reportUnmatched: false
   treatPhpDocTypesAsCertain: false
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -1,8 +1,7 @@
 parameters:
   ignoreErrors:
     -
-      message: '#@throws with type .+SimpleCache.+ is not subtype of Throwable#'
-      path: src/Repository/DefaultUnleashRepository.php
+      message: '#@throws with type .+ is not subtype of Throwable#'
       reportUnmatched: false
     -
       messages:
--- a/tests/Traits/FakeCacheImplementationTrait.php
+++ b/tests/Traits/FakeCacheImplementationTrait.php
@@ -3,51 +3,13 @@
 namespace Unleash\Client\Tests\Traits;
 
 use Psr\SimpleCache\CacheInterface;
+use Symfony\Component\Cache\Adapter\NullAdapter;
+use Symfony\Component\Cache\Psr16Cache;
 
 trait FakeCacheImplementationTrait
 {
     private function getCache(): CacheInterface
     {
-        return new class implements CacheInterface {
-            public function get($key, $default = null)
-            {
-                return $default;
-            }
-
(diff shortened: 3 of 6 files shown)


.github/workflows/tests.yaml
```diff
@@ -58,7 +60,11 @@ jobs:
         uses: actions/checkout@v2
         with:
           submodules: true
-      - name: Install Dependencies
+      - name: Install Dependencies (without optional)
+        if: ${{ matrix.optional-deps == '' }}
         run: composer install
+      - name: Install Dependencies (with optional)
+        if: ${{ matrix.optional-deps != '' }}
+        run: composer require ${{ matrix.optional-deps }}
```
colinodell, line 68: The matrix will allow us to run these tests twice:

 - Once without `cache/filesystem-adapter` (and therefore using `psr/simple-cache` 3.x)
 - Once with `cache/filesystem-adapter` (and therefore using `psr/simple-cache` 1.x)

Thus giving us confidence that everything works as expected.

phpstan.neon.dist
```diff
@@ -1,4 +1,13 @@
...
+    -
+      message: '#@throws with type .+SimpleCache.+ is not subtype of Throwable#'
+      path: src/Repository/DefaultUnleashRepository.php
+      reportUnmatched: false
+    -
+      messages:
+        - '#Class .+Filesystem.* not found#'
+        - '#Class .+Flysystem.* not found#'
+      path: src/Helper/DefaultImplementationLocator.php
+      reportUnmatched: false
```
colinodell, line 11: These message will be reported if `cache/filesystem-adapter` is missing, and that's okay.

tests/Traits/FakeCacheImplementationTrait.php
```diff
@@ -3,51 +3,13 @@
...
-            {
-                return true;
-            }
-
-            public function has($key): bool
-            {
-                return false;
-            }
-        };
+        return new Psr16Cache(new NullAdapter());
```
colinodell, line 13: `symfony/cache` is already required as a dev dependency, so we can use it's `NullAdapter` which does exactly the same thing, and it gracefully supports `psr/simple-cache` the varied types used in 1.x and 3.x!

tests/Traits/RealCacheImplementationTrait.php
```diff
@@ -24,11 +23,7 @@ protected function tearDown(): void
     private function getCache(): CacheInterface
     {
         if ($this->_cache === null) {
-            $this->_cache = new FilesystemCachePool(
-                new Filesystem(
-                    new Local(sys_get_temp_dir() . '/unleash-sdk-tests')
-                )
-            );
+            $this->_cache = new Psr16Cache(new ArrayAdapter());
```
colinodell, line 26: Similarly, we can use `symfony/cache` here too, which is especially useful now that `cache/filesystem-adapter` may not be installed when the test suite runs.

tests/UnleashBuilderTest.php
```diff
@@ -339,28 +339,36 @@ public function testWithoutDefaultClients()
...
-
         $defaultImplementationsProperty = (new ReflectionObject($locator))->getProperty('defaultImplementations');
         $defaultImplementationsProperty->setAccessible(true);
         $defaultImplementations = $defaultImplementationsProperty->getValue($locator);
 
-        $defaultImplementations['cache'][FilesystemCachePool::class . 2] = [];
-        unset($defaultImplementations['cache'][FilesystemCachePool::class]);
-        $defaultImplementationsProperty->setValue($locator, $defaultImplementations);
+        // cache/filesystem-adapter should be used by default if it's installed
+        if (class_exists(FilesystemCachePool::class)) {
```
colinodell, line 347: This looks like a lot of changes, but it's just rearranging existing code and only running `cache/filesystem-adapter`-related tests if it's installed.

colinodell (commented):
Some notes to assist with your review :)

sighphyre:
@colinodell This looks great! Thanks for doing this!

I'm happy with this but I'd like @RikudouSage to take a glance and give the nod

phpstan.neon.dist
RikudouSage: @colinodell Can you put back the original version of throwing not Throwable? I'd like to ignore that error globally, because it' stupid.

RikudouSage (changes requested):
This looks great! There's just one small thing and then I'm more than happy to merge!

phpstan.neon.dist
colinodell: Sure thing! Fixed via 4121a65.

RikudouSage (approved):

PR #123, opened 2022-10-28
Read the response body from the start

### LINKED ISSUE #122: Bug: DefaultUnleashRepository fails to read entire response body

### Describe the bug

The `DefaultUnleashRepository` uses this code to read the response body:

https://github.com/Unleash/unleash-client-php/blob/2e2e66f5b341f8bdaea5f98ba9761658a5346a67/src/Repository/DefaultUnleashRepository.php#L90-L94

The `getContents()` method [reads "the **remaining** contents" of the response body](https://github.com/php-fig/http-message/blob/efd67d1dc14a7ef4fc4e518e7dee91c271d524e4/src/StreamInterface.php#L136-L143).  This causes issues when the user supplies a Guzzle instance with middleware that also reads the response body, because the middleware will have read the body through the end of the stream, thus setting the current position at the end of the stream so that all future calls to `getContents()` will return an empty string.

As a result, the current code to read the response body may produce an empty string which then causes a `JsonException : Syntax error` later on.

### To reproduce

Create an Unleash instance with a user-supplied Guzzle client containing any middleware that reads the response body.  It could be Guzzle's `Middleware::log()` or even something as simple as this:

```php
// Guzzle's `mapResponse()` helper is used to create middleware that manipulates responses
Middleware::mapResponse(
    static function ($response) {
        // Simply read the response body, causing the stream to seek to the end, and return the response as-is
        $response->getBody()->getContents();
        return $response;
    }
)
```

### Sample code (optional)

_No response_

### Version

v1.8.081

### Expected behavior

The code should always read the stream from the start because it's possible something else has read the stream first

### Logs (optional)

_No response_

I plan to submit a PR to fix this shortly :)

colinodell:
# Description

Fixes #122 by using `__toString()` to read the response body [which always attempts to read the stream from the start](https://github.com/php-fig/http-message/blob/efd67d1dc14a7ef4fc4e518e7dee91c271d524e4/src/StreamInterface.php#L14-L28). 

## Type of change

# How Has This Been Tested?

- [x] Unit tests
- [ ] Spec Tests

# Checklist:

### DIFF
--- a/src/Repository/DefaultUnleashRepository.php
+++ b/src/Repository/DefaultUnleashRepository.php
@@ -89,7 +89,7 @@ final class DefaultUnleashRepository implements UnleashRepository
                 try {
                     $response = $this->httpClient->sendRequest($request);
                     if ($response->getStatusCode() === 200) {
-                        $data = $response->getBody()->getContents();
+                        $data = (string) $response->getBody();
                         $this->setLastValidState($data);
                     }
                 } catch (Exception $exception) {
--- a/tests/Repository/DefaultUnleashRepositoryTest.php
+++ b/tests/Repository/DefaultUnleashRepositoryTest.php
@@ -4,9 +4,11 @@ namespace Unleash\Client\Tests\Repository;
 
 use GuzzleHttp\Client;
 use GuzzleHttp\HandlerStack;
+use GuzzleHttp\Middleware;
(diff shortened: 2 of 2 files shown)


RikudouSage:
Thank you very much!

RikudouSage (approved):

PR #130, opened 2022-12-09
Fix: Load stale data whenever there's any http error

### LINKED ISSUE #129: Bug: stale data cache is not loaded when HTTP communication fails with 4xx or 5xx codes

### Describe the bug

The docs state:

> Stale data cache is used when http communication fails while fetching feature list from the server.

The DefaultUnleashRepository only fetches stale cache when an exception is thrown:

```php
                try {
                    $response = $this->httpClient->sendRequest($request);
                    if ($response->getStatusCode() === 200) {
                        $data = $response->getBody()->getContents();
                        $this->setLastValidState($data);
                    }
                } catch (Exception $exception) {
                    $this->configuration->getEventDispatcherOrNull()?->dispatch(
                        new FetchingDataFailedEvent($exception),
                        UnleashEvents::FETCHING_DATA_FAILED,
                    );
                    $data = $this->getLastValidState();
                }
```

If a 4xx or 5xx code is returned, the `FETCHING_DATA_FAILED` event is not dispatched, and last valid state not fetched.

The Client's `sendRequest()` method implements PSR-7, which expects the method to return instead of throwing in the event of a valid response but 4xx or 5xx error code. Since Unleash fails to fetch data from the server with these header codes (according to HTTP conventions), this should be fixed.

My hacky workaround was to create a new Client child class that forces it to throw on non-2xx header codes. Setting the `http_errors` option on the default client does not work.

### To reproduce

1. Set up the code to use the default Unleash repository with a cache
2. Make a request to get a response and put data in the cace
3. Change the authorization in the configuration to something incorrect
4. Make another request (this will cause the HTTP request to return a 401 header code)
5. An exception is thrown

### Sample code (optional)

_No response_

### Version

v1.8.081

### Expected behavior

Stale cache should be loaded, but it's not.

### Logs (optional)

_No response_

_No response_

RikudouSage:
# Description

Fixes incorrect behavior when stale fallback data weren't returned when non-200 status code was returned.

Fixes #129 

## Type of change

# How Has This Been Tested?

- [x] Unit tests
- [x] Integration tests

# Checklist:

### DIFF
--- a/src/Repository/DefaultUnleashRepository.php
+++ b/src/Repository/DefaultUnleashRepository.php
@@ -91,6 +91,8 @@ final class DefaultUnleashRepository implements UnleashRepository
                     if ($response->getStatusCode() === 200) {
                         $data = (string) $response->getBody();
                         $this->setLastValidState($data);
+                    } else {
+                        throw new HttpResponseException("Invalid status code: '{$response->getStatusCode()}'");
                     }
                 } catch (Exception $exception) {
                     $this->configuration->getEventDispatcherOrNull()?->dispatch(
--- a/tests/Repository/DefaultUnleashRepositoryTest.php
+++ b/tests/Repository/DefaultUnleashRepositoryTest.php
@@ -349,4 +349,39 @@ final class DefaultUnleashRepositoryTest extends AbstractHttpClientTest
             sleep(1);
         }
     }
+
+    /**
+     * @see https://github.com/Unleash/unleash-client-php/issues/129
+     */
+    public function testFallbackStaleCacheNoException()
+    {
+        $eventEmittedCount = 0;
+
+        $eventDispatcher = new EventDispatcher(new SymfonyEventDispatcher());
+        $eventDispatcher->addListener(
+            UnleashEvents::FETCHING_DATA_FAILED,
+            function (FetchingDataFailedEvent $event) use (&$eventEmittedCount): void {
+                $event->getException(); // just to cover the line
+                ++$eventEmittedCount;
+            }
+        );
+
+        $repository = new DefaultUnleashRepository(
+            new Client([
+                'handler' => $this->handlerStack,
+            ]),
+            new HttpFactory(),
+            (new UnleashConfiguration('', '', ''))
+                ->setCache($this->getRealCache())
+                ->setEventDispatcher($eventDispatcher)
+                ->setTtl(0)
+                ->setStaleTtl(3)
+        );
+
+        $this->pushResponse($this->response);
+        $this->pushResponse($this->response, 1, 401);
+        $features = $repository->getFeatures();
+        self::assertEquals($features, $repository->getFeatures());
+        self::assertSame(1, $eventEmittedCount);
+    }
 }


src/Repository/DefaultUnleashRepository.php
```diff
@@ -91,6 +91,8 @@ public function getFeatures(): iterable
                     if ($response->getStatusCode() === 200) {
                         $data = (string) $response->getBody();
                         $this->setLastValidState($data);
+                    } else {
+                        throw new HttpResponseException("Invalid status code: '{$response->getStatusCode()}'");
```
sighphyre, line 95: I think this is correct. I don't believe PHP is ever going to do the HTTP 304 dance like the other SDKs and anything in the 500 range should be fine to get handled in the catch.

There's two 400 range HTTP status codes I want to clarify:
401 and 429. Do we want to raise error events in the dispatcher when the client is either being denied due to a bad token or rate limited by our infra? I want to say yes, just a sanity check that you've thought about it and agree

RikudouSage: 304 should be handled automatically and shouldn't be any issue.

429 seems like the perfect use-case for this, I'm a little conflicted on the 401, but ultimately I think that the SDK should try to fail as little as possible.

There is a time limit for the stale cache, meaning if the 401 prevails it will fail with an exception eventually, while temporary outages somewhere along the line (like some proxy having an outage) which could cause the 401 will be handled gracefully.

eblanshey: > ultimately I think that the SDK should try to fail as little as possible.

Agree with this.

Not critical, but should they body also be supplied in the exception message to help devs debug errors, in case it's available?

RikudouSage: @eblanshey I don't think it would be useful given the exception is caught immediately so nothing ever propagates further. Or did you mean somewhere else?

sighphyre: >429 seems like the perfect use-case for this, I'm a little conflicted on the 401, but ultimately I think that the SDK should try to fail as little as possible.

Definitely don't think this should cause a failure, idly wondering about the value of being able to respond to a 401 with a hook. Don't think it's harmful to have though so yeah, looks good

sighphyre (approved):
LGTM

src/Repository/DefaultUnleashRepository.php
sighphyre: >There is a time limit for the stale cache, meaning if the 401 prevails it will fail with an exception eventually, while temporary outages somewhere along the line (like some proxy having an outage) which could cause the 401 will be handled gracefully.

Not related to this PR so maybe a conversation for a different time and place but is this what we want? My gut feel is the SDK raising an exception almost anywhere is surprising and not really handleable by the caller. Wouldn't mind unpacking this at some point (honestly I'd like a standard set of behaviours for our SDKs on this, which we don't really have right now)

eblanshey: > I don't think it would be useful given the exception is caught immediately so nothing ever propagates further. Or did you mean somewhere else?

The exception is passed to the `FetchingDataFailedEvent`, so I'd hook into that event to add a message to our logs. The body would help as much as the header code for context. For example, it could say "Incorrect API key", or "API key expired" or "Account banned", to name a few examples off the top of my head.

But I see this was already merged so I guess we'll leave it at that.

PR #135, opened 2022-12-12
Feat: Add separate stale cache handler config

### LINKED ISSUE #131: Feat: allow separate cache driver to be used for stale cache

### Describe the new feature

There are two levels of caching currently: the 30s short cache, and the stale cache which is longer. Additionally, there is a way to provide a bootstrap file that can be used in case a request to the server fails. So in essence there are 3 ways of falling back to something if a request fails.

A bootstrap is persistent, whereas the caches are temporary. What I would like is to have a more persistent "stale" cache, that stores the result of the latest known request. This is better than a bootstrap file that needs be manually updated with the latest feature flags.

The stale cache that's currently implemented works well, but uses the same cache provider as the short term 30s cache. The two usages are quite different: one is for frequent requests in a cache store that is ephemeral (like Memcached and Redis with automatic eviction.) The stale cache is more long term and shouldn't be wiped out automatically with evictions, new deployments, etc -- it is used to prevent errors from bubbling up on production.

It would be nice to provide the ability to set a separate cache driver for the stale cache, via the builder, such as with a `withStaleCacheHandler()` method. The stale cache handler can then use a more persistent storage such as a file or persistent Redis instance. Consider it like a way to auto-populate a bootstrap file.

To make it backwards compatible with previous releases, the stale cache can use the regular cache provider if one is not provided.

_No response_

_No response_

_No response_

RikudouSage:
# Description

Includes a configuration to set different stale cache handler.

Fixes #131 

## Type of change

# How Has This Been Tested?

- [x] Unit tests
- [x] Integration tests

# Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -264,6 +264,31 @@ $builder = $builder
     ->withCacheHandler(null);
 ```
 
+You can use a different cache implementation for standard item cache and for stale cache. If you don't provide any
+implementation for stale cache, the same instance as for standard cache is used.
+
+```php
+<?php
+
+use Cache\Adapter\Filesystem\FilesystemCachePool;
+use League\Flysystem\Adapter\Local;
+use League\Flysystem\Filesystem;
+use Unleash\Client\UnleashBuilder;
+use Symfony\Component\Cache\Psr16Cache;
+use Symfony\Component\Cache\Adapter\ArrayAdapter;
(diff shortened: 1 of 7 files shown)


sighphyre (approved):
LGTM

PR #143, opened 2023-01-25
Fix/ends with operator

sighphyre:
This patches the strEndsWith operator to check that constraint property ends with the context property rather than the reverse. The reason the client specifications were passing is because of a bug in that project (patched here: https://github.com/Unleash/client-specification/pull/47).

strStartsWith is not affected, the check is correct here: https://github.com/Unleash/unleash-client-php/blob/main/src/ConstraintValidator/Operator/String/StringStartsWithOperatorValidator.php

We might want to think about what version we release this under. Technically it's a bug fix but there's a risk of having an automated bot upgrade the SDK because it sees a patch version and flipping a bunch of toggles in the wild.

### DIFF
--- a/src/ConstraintValidator/Operator/String/StringEndsWithOperatorValidator.php
+++ b/src/ConstraintValidator/Operator/String/StringEndsWithOperatorValidator.php
@@ -11,6 +11,6 @@ final class StringEndsWithOperatorValidator extends AbstractStringOperatorValida
     {
         assert(is_string($searchInValue));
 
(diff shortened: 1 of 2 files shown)


ivarconr (approved):

ivarconr:
> Technically it's a bug fix but there's a risk of having an automated bot upgrade the SDK because it sees a patch version and flipping a bunch of toggles in the wild.

I hear you, at the same time this is a bug breaking expectations on how Unleash works. I would not be too worried as the intention is most likely it should be enabled in this case.

PR #150, opened 2023-04-27
Add support for psr/http-message:^2.0

W0rma:
# Description

[Release notes for `psr/http-message:^2.0`](https://github.com/php-fig/http-message/releases/tag/2.0)

## Type of change

- [x] Dependency update

# How Has This Been Tested?

- [x] Unit tests
- [ ] Spec Tests

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
         "psr/simple-cache": "^1.0 | ^2.0 | ^3.0",
         "psr/simple-cache-implementation": "^1.0 | ^2.0 | ^3.0",
         "lastguest/murmurhash": "^2.1",
-        "psr/http-message": "^1.0",
+        "psr/http-message": "^1.0 | ^2.0",
         "php-http/discovery": "^1.14"
     },
     "autoload": {
--- a/src/Helper/StringStream.php
+++ b/src/Helper/StringStream.php
@@ -32,7 +32,7 @@ final class StringStream implements StreamInterface
         rewind($this->stream);
     }
 
-    public function __toString()
+    public function __toString(): string
     {
         $stream = $this->stream;
         if ($stream === null) {


RikudouSage:
It doesn't seem to pass tests.

W0rma:
> It doesn't seem to pass tests.

Sorry, there was a typo.

Should be fixed now. I was able to run the test suite locally.

RikudouSage:
Can you make it accept ^1.0 as well? Also remove the changes from StringStream, I feel like it adds unnecessary BC break, the only change that's needed is adding `string` return type to `__toString()`.

W0rma:
> Can you make it accept ^1.0 as well? Also remove the changes from StringStream

I've just removed my changes.

RikudouSage:
As mentioned before, the only change that's needed is adding string return type to __toString(), otherwise it doesn't pass. If you rebase the current main branch, additional tests would be ran that should catch the error.

Andreas-Halemba:
Are here any updates on this?

W0rma:
> As mentioned before, the only change that's needed is adding string return type to __toString(), otherwise it doesn't pass.

@RikudouSage I have just rebased and the tests are still green.

I don't see how the requested change should be related to the changeset of this MR.

Please let me know if you still want me to change this method signature.

RikudouSage:
Hm, that's weird, I'll check it out today.

RikudouSage:
Hm, it seems to work because PHP internally adds `string` return type to the `__toString()` method, but I'd like it explicit anyway, because as of now it doesn't follow Liskov substitution principle, even though PHP changes it internally to be compatible.

I'm talking about the method `\Unleash\Client\Helper\StringStream::__toString()`.

W0rma:
@RikudouSage I just added the return type as suggested.

RikudouSage (approved):
Thanks!

PR #152, opened 2023-05-19
Fix: Single value constraints

### LINKED ISSUE #151: Bug: Constraints check doesn't work

### Describe the bug

Constraints checks don't work. The issue is in DefaultConstraintsValidator.php on line 33:
```
$valueToPass = $constraint->getValues() ?? (
            method_exists($constraint, 'getSingleValue')
                ? $constraint->getSingleValue()
                : null
        );
```

This code doesn't work as intended because `$constraints->getValues()` returns an empty array. The null coalescing operator only works for null values, not for an empty array value. 

The code could be rewritten like this (to fix the issue):
```
$valueToPass = $constraint->getValues();
if (!$valueToPass) {
    $valueToPass = $constraint->getSingleValue() ?? null;
}
```

### To reproduce

This can be reproduced by using a DATE_BEFORE constraint on the currentTime. Just add the constraint on the default strategy. The feature toggle will evaluate to "off" as soon as you add the constraint.

### Sample code (optional)

_No response_

### Version

v1.11.082

### Expected behavior

I expect the constraint to be properly evaluated. Instead it always evaluates to false.

### Logs (optional)

_No response_

I'm using unleash-client-php in a Laravel project, together with the j-webb/unleash package.

RikudouSage:
# Description

Fixes a bug where single value constraints weren't evaluated at all.

Fixes #151 

## Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/src/ConstraintValidator/DefaultConstraintValidator.php
+++ b/src/ConstraintValidator/DefaultConstraintValidator.php
@@ -30,11 +30,9 @@ final class DefaultConstraintValidator implements ConstraintValidator
         $currentValue = $context->findContextValue($field) ?? '';
 
         $callback = $this->getValidationCallback($constraint->getOperator());
-        $valueToPass = $constraint->getValues() ?? (
-            method_exists($constraint, 'getSingleValue')
-                ? $constraint->getSingleValue()
-                : null
-        );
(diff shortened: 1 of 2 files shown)


sighphyre (approved):
Awesome! Thanks for looking into this one. LGTM!

PR #156, opened 2023-05-22
Chore: Ignore rector configuration file in dist packages

RikudouSage:
# Description

Ignores rector configuration file in dist packages as it's not needed if installed as a dependency.


sighphyre (approved):
LGTM

PR #159, opened 2023-05-23
Fix: Code style error after updating code style fixer

RikudouSage:
# Description

There was some code that wasn't formatted correctly and the fixer didn't catch the error previously, now it does.


sighphyre (approved):
LGTM!

PR #161, opened 2023-05-25
Added getFeatures and getFeature functions in DefaultUnleash Class and written test cases for these function in DefaultUnleashTest

bs084s:
# Description

Please include a summary of the change and which issue is fixed (if any).
We have create two custom functions getFeatures and getFeature into DefaultUnleash Class. In our application we need list of toggle that why we have create these functions. 

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [x] Unit tests
- [ ] Spec Tests

# Checklist:

### DIFF
--- a/src/Unleash.php
+++ b/src/Unleash.php
@@ -14,4 +14,8 @@ interface Unleash
     public function getVariant(string $featureName, ?Context $context = null, ?Variant $fallbackVariant = null): Variant;
 
     public function register(): bool;
+
+    public function getFeatures(): array;
+
+    public function getFeature(string $featureName);
 }
--- a/src/Unleash.php
+++ b/src/Unleash.php
@@ -14,8 +14,4 @@ interface Unleash
     public function getVariant(string $featureName, ?Context $context = null, ?Variant $fallbackVariant = null): Variant;
 
     public function register(): bool;
-
-    public function getFeatures(): array;
-
-    public function getFeature(string $featureName);
 }
--- a/tests/DefaultUnleashTest.php
+++ b/tests/DefaultUnleashTest.php
@@ -826,4 +826,142 @@ final class DefaultUnleashTest extends AbstractHttpClientTest
             new DefaultVariantHandler(new MurmurHashCalculator())
         );
     }
+
+    public function testGetFeatures()
+    {
+        $instance = $this->getInstance(new DefaultStrategyHandler());
+
+        $this->pushResponse([
+            'version' => 1,
+            'features' => [
+                [
+                    'name' => 'test',
+                    'description' => '',
+                    'enabled' => false,
+                    'impressionData' => true,
+                    'strategies' => [
+                        [
+                            'name' => 'default',
+                        ],
+                    ],
+                ],
+                [
+                    'name' => 'test2',
+                    'description' => '',
+                    'enabled' => true,
+                    'impressionData' => true,
+                    'strategies' => [
+                        [
+                            'name' => 'default',
+                        ],
+                    ],
+                    'variants' => [
+                        [
+                            'name' => 'variant1',
+                            'weight' => 1,
+                            'payload' => [
+                                'type' => 'string',
+                                'value' => 'val1',
+                            ],
+                        ],
+                    ],
+                ],
+            ],
+        ]);
+
+        self::assertNotNull($instance->getFeatures());
+    }
+
+    public function testGetFeaturesEmpty()
+    {
+        $instance = $this->getInstance(new DefaultStrategyHandler());
+
+        $this->pushResponse([
+            'version' => 1,
+            'features' => [],
+        ]);
+
+        self::assertEmpty($instance->getFeatures());
+    }
+
+    public function testGetFeaturesCount()
+    {
+        $instance = $this->getInstance(new DefaultStrategyHandler());
+
+        $this->pushResponse([
+            'version' => 1,
+            'features' => [
+                [
+                    'name' => 'test',
+                    'description' => '',
+                    'enabled' => false,
+                    'impressionData' => true,
+                    'strategies' => [
+                        [
+                            'name' => 'default',
+                        ],
+                    ],
+                ],
+                [
+                    'name' => 'test2',
+                    'description' => '',
+                    'enabled' => true,
+                    'impressionData' => true,
+                    'strategies' => [
+                        [
+                            'name' => 'default',
+                        ],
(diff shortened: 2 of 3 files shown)


src/Unleash.php
```diff
@@ -14,4 +14,8 @@ public function isEnabled(string $featureName, ?Context $context = null, bool $d
     public function getVariant(string $featureName, ?Context $context = null, ?Variant $fallbackVariant = null): Variant;
 
     public function register(): bool;
+
+    public function getFeatures(): array;
+
+    public function getFeature(string $featureName);
```
RikudouSage: We cannot add these two methods to the interface for now, it would be a breaking change (meaning anyone who was already implementing the interface would suddenly get fatal errors if they updated the package).

What should be done is adding these as soft requirement - meaning only as a docblock comment as for example here: https://github.com/Unleash/unleash-client-php/blob/main/src/Configuration/Context.php#L10

That way the methods are not enforced but people have time to add them in their implementations. When next major version is released, these soft-enforced methods will be moved to the real methods.

src/Unleash.php
```diff
@@ -14,4 +14,8 @@ public function isEnabled(string $featureName, ?Context $context = null, bool $d
     public function getVariant(string $featureName, ?Context $context = null, ?Variant $fallbackVariant = null): Variant;
 
     public function register(): bool;
+
+    public function getFeatures(): array;
```
RikudouSage: It's missing the type it returns, it should be added in a docblock like this: https://github.com/Unleash/unleash-client-php/blob/main/src/Repository/UnleashRepository.php#L12

src/Unleash.php
RikudouSage: This doesn't have a return type, it should have one: https://github.com/Unleash/unleash-client-php/blob/main/src/Repository/UnleashRepository.php#L9

tests/DefaultUnleashTest.php
```diff
@@ -826,4 +826,142 @@ private function getInstance(StrategyHandler ...$handlers): DefaultUnleash
...
+                                'type' => 'string',
+                                'value' => 'val1',
+                            ],
+                        ],
+                    ],
+                ],
+            ],
+        ]);
+
+        self::assertNotNull($instance->getFeatures());
```
RikudouSage: This doesn't make sense, the method can never return null, you should test that the count of features is two and test that individual features are correctly returned.

src/Unleash.php
RikudouSage: Given this is a proxy method to the `findFeature()` method in repository, it should be named the same.

RikudouSage (changes requested):
I think we've talked on Slack?

Anyway, to help you out I've added some comments, even though this PR is not at all a direction I'd like this to go.

Take those comments as a general guidelines to help you out in general development.

I'll have to think if we even want to expose the object like that, but if so it would be done by exposing the repository on its own, meaning adding a single `getRepository()` method. Like this you'd be basically needlessly creating proxy methods for the repository object.

I'll get back to you with information on how to proceed regarding exposing the `getRepository()` method if you'd want to change the PR to do that instead.

bs084s:
@RikudouSage I have fixed all the issues that you mentioned and also resolved the static analysis and code style issues.

RikudouSage:
Hello,

as mentioned earlier, this is not the way we'd like this to go.

If you need to get the list of features, you currently have to construct the object manually, I've just added an example in https://github.com/Unleash/unleash-client-php/pull/162, but for the sake of completeness, it's also here:

```php
<?php

use Http\Discovery\Psr17FactoryDiscovery;
use Http\Discovery\Psr18ClientDiscovery;
use Unleash\Client\Client\DefaultRegistrationService;
use Unleash\Client\Configuration\UnleashConfiguration;
use Unleash\Client\DefaultUnleash;
use Unleash\Client\Metrics\DefaultMetricsHandler;
use Unleash\Client\Metrics\DefaultMetricsSender;
use Unleash\Client\Repository\DefaultUnleashRepository;
use Unleash\Client\Stickiness\MurmurHashCalculator;
use Unleash\Client\Strategy\ApplicationHostnameStrategyHandler;
use Unleash\Client\Strategy\DefaultStrategyHandler;
use Unleash\Client\Strategy\GradualRolloutRandomStrategyHandler;
use Unleash\Client\Strategy\GradualRolloutSessionIdStrategyHandler;
use Unleash\Client\Strategy\GradualRolloutStrategyHandler;
use Unleash\Client\Strategy\GradualRolloutUserIdStrategyHandler;
use Unleash\Client\Strategy\IpAddressStrategyHandler;
use Unleash\Client\Strategy\UserIdStrategyHandler;
use Unleash\Client\Variant\DefaultVariantHandler;

$stickinessCalculator = new MurmurHashCalculator();
$gradualRolloutStrategyHandler = new GradualRolloutStrategyHandler($stickinessCalculator);
$strategies = [
    $gradualRolloutStrategyHandler,
    new ApplicationHostnameStrategyHandler(),
    new DefaultStrategyHandler(),
    new IpAddressStrategyHandler(),
    new UserIdStrategyHandler(),
    // you can skip these three if you don't use the deprecated strategies
    new GradualRolloutRandomStrategyHandler($gradualRolloutStrategyHandler),
    new GradualRolloutSessionIdStrategyHandler($gradualRolloutStrategyHandler),
    new GradualRolloutUserIdStrategyHandler($gradualRolloutStrategyHandler),
];
$httpClient = Psr18ClientDiscovery::find();
$httpFactory = Psr17FactoryDiscovery::findRequestFactory();
$configuration = (new UnleashConfiguration(
    $appUrl,
    $appName,
    $instanceId,
))->setHeaders([
    'Authorization' => $apiKey,
]);
// add any other configuration to the configuration object, like cache configuration etc.
$repository = new DefaultUnleashRepository($httpClient, $httpFactory, $configuration);
$registrationService = new DefaultRegistrationService($httpClient, $httpFactory, $configuration);
$metricsSender = new DefaultMetricsSender($httpClient, $httpFactory, $configuration);
$metricsHandler = new DefaultMetricsHandler($metricsSender, $configuration);
$variantHandler = new DefaultVariantHandler($stickinessCalculator);

$unleash = new DefaultUnleash(
    $strategies,
    $repository,
    $registrationService,
    $configuration,
    $metricsHandler,
    $variantHandler,
);

```

That way you have access to the repository object in the `$repository` variable and can call `getFeatures()` on it.

PR #163, opened 2023-06-07
Fix: Catch an exception when a http failure occurs

RikudouSage:
# Description

Fixes an error reported on Slack where a http exception isn't caught.

## Type of change

# How Has This Been Tested?

- [x] Unit tests

# Checklist:

### DIFF
--- a/src/Metrics/DefaultMetricsSender.php
+++ b/src/Metrics/DefaultMetricsSender.php
@@ -2,6 +2,7 @@
 
 namespace Unleash\Client\Metrics;
 
+use Psr\Http\Client\ClientExceptionInterface;
 use Psr\Http\Client\ClientInterface;
 use Psr\Http\Message\RequestFactoryInterface;
 use Unleash\Client\Configuration\UnleashConfiguration;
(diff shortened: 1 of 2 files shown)


src/Metrics/DefaultMetricsSender.php
```diff
@@ -33,6 +34,11 @@ public function sendMetrics(MetricsBucket $bucket): void
         foreach ($this->configuration->getHeaders() as $name => $value) {
             $request = $request->withHeader($name, $value);
         }
-        $this->httpClient->sendRequest($request);
+
+        try {
+            $this->httpClient->sendRequest($request);
+        } catch (ClientExceptionInterface) {
+            // ignore the error
```
sighphyre, line 41: This looks intentionally ignored but do we not want to log this? Totally okay with the answer being no here

sighphyre (approved):
LGTM!

src/Metrics/DefaultMetricsSender.php
RikudouSage: It doesn't currently log anything as there's no logging set up, not sure how to log this, the only thing I can come up with is [`error_log()`](https://www.php.net/manual/en/function.error-log.php). The SDK would probably need some logger support in general, not sure it's worth creating a partial support here.

sighphyre: No sounds super sensible to me!

PR #165, opened 2023-07-12
feat: proxy sdk

sighphyre:
This adds implementation of `Unleash` that talks to the proxy endpoints on Edge. This implements the same external interface as the existing `DefaultUnleash` but the implementation details of a Proxy SDK are very different behind the scenes, which makes for some significant design decisions. The responses returned from the proxy endpoints do not return enough data for the existing SDK to reuse so I've split this out at the repository layer. I don't believe this is a major problem, the construction of the objects is handled by the existing `UnleashBuilder`, consumers that want to manually construct instances will need to know what they're doing.

- The API responses are completely different, toggles in Client response can't be handled independently because of segments, while the Proxy responses can. Because of this, I've used the single toggle responses for the Proxy SDK so that the consumer will only incur the cost of toggles they use (API and cache throughput). This means retrieving from cache is significantly lighter than using the existing `DefaultUnleash`
- Registration is not handled by Proxy SDKs directly, this is handled by the Proxy or Edge and so registration is a no-op in the `DefaultProxyUnleash` class
- The `DefaultProxyUnleash` class reuses the existing builder, context, caching layers, configuration, and metrics handlers. It does **not** reuse the existing `UnleashRepository` or `Feature` classes. This is intentional for a few reasons: behind the scenes retrieving the relevant feature toggle for the Proxy SDK requires the actual context object to make the decision on how to resolve the feature (the resolution itself is handled by Edge/Proxy). So the method signature looks like this:

```
    public function findFeatureByContext(string $featureName, ?Context $context = null): ?ProxyFeature;

```
- The above is pretty much unavoidable, since that's how the Edge/Proxy works. The returned `ProxyFeature` is a very stripped down version of whats contained in the existing `Feature` class, but more to the point, the fields mean vastly different things. While it's probably possible to force the `DefaultProxyFeature` and `DefaultFeature` classes to share an interface, I don't believe this is correct, since doing so will signal to consumers that they're interchangeable, when they very much are not.
- Because the variant returned from Edge/Proxy contains only the fields that are useful for the consumer and not things like `weight` or `stickiness` I've created a new interface and class called `ResolvedVariant`/`DefaultResolvedVariant` which is now returned from the high level public interfaces, containing only the relevant fields. The existing code now uses the existing `Variant`/`DefaultVariant` for internal calculations but no longer exposes it through the high level interfaces. **This is a breaking change, albeit a small one**


RikudouSage:
I've only skimmed through this, but IMO we should use the existing interfaces more, for example the `register()` method can be noop in the proxy.

sighphyre:
> I've only skimmed through this, but IMO we should use the existing interfaces more, for example the `register()` method can be noop in the proxy.

Yeah, I think that's probably a good plan. I actually had them implement the same interface right until the point I added the `getVariant` method which is the problem here. The variant object returned by that method is currently exposing things like weight, stickiness and overrides, which we never receive from the proxy/edge/frontendapi endpoints. Those fields are really internal implementation details and I don't think they're useful to the consumer. If we trim that interface down to only return enabled, name and payload then it's easy enough to have both ProxyUnleash and Unleash use the same interface but it does require a breaking change. WDYT?

examples/09-cached-proxy.php
```diff
@@ -0,0 +1,41 @@
...
+
+require __DIR__ . '/_common.php';
+
+use Unleash\Client\UnleashBuilder;
+use Symfony\Component\Cache\Adapter\FilesystemAdapter;
+use Symfony\Component\Cache\Psr16Cache;
+
+
+$psr6Cache = new FilesystemAdapter();
+$cache = new Psr16Cache($psr6Cache);
```
gastonfournier, line 11: Nit:
```suggestion
$psr16Cache = new FilesystemAdapter();
$cache = new Psr16Cache($psr16Cache);
```

src/DefaultUnleash.php
```diff
@@ -115,14 +116,14 @@ public function isEnabled(string $featureName, ?Context $context = null, bool $d
...
         $fallbackVariant ??= $this->variantHandler->getDefaultVariant();
         $context ??= $this->configuration->getContextProvider()->getContext();
 
         $feature = $this->repository->findFeature($featureName);
         if ($feature === null || !$feature->isEnabled() || !count($feature->getVariants())) {
-            return $fallbackVariant;
+            return new DefaultResolvedVariant($fallbackVariant->getName(), $fallbackVariant->isEnabled(), $fallbackVariant->getPayload());
         }
 
         $variant = $this->variantHandler->selectVariant($feature, $context);
```
gastonfournier: Maybe this can be refactored so this https://github.com/Unleash/unleash-client-php/pull/165/files#diff-e4aabb8061e2fa61c3253aea34f8d4478c027fccc9fd63d2953e09635f885b2cR145 always executes. It might be worthy, just for maintainability:
```php
        $feature = $this->repository->findFeature($featureName);
        if ($feature === null || !$feature->isEnabled() || !count($feature->getVariants())) {
            $variant = null;
        } else {
            $variant = $this->variantHandler->selectVariant($feature, $context);
        }
```

examples/09-cached-proxy.php
```diff
@@ -0,0 +1,41 @@
...
+    echo "myFeature is enabled \n";
+} else {
+    echo "myFeature is disabled \n";
+}
+
+$resolvedVariant = $unleash->getVariant('myFeature');
+echo "Resolved a variant";
+var_dump($resolvedVariant);
+
+sleep(1);
```
gastonfournier, line 35: Is this needed? If yes can we add a comment on why?

src/Repository/DefaultUnleashProxyRepository.php
```diff
@@ -0,0 +1,207 @@
...
+            if (is_array($cachedFeature)) {
+                $featureData = $this->validateResponse($cachedFeature);
+                if ($featureData !== null) {
+                    return new DefaultProxyFeature($featureData);
+                }
+            }
+        }
+
+        $context ??= new UnleashContext();
+        $featureUrl = $this->configuration->getUrl() . 'frontend/features/' . $featureName;
```
gastonfournier, line 46: Just because I don't know the answer: Do we need to url-encode `$featureName`? I think this is the first time we add a feature name in a URL: https://sourcegraph.com/search?q=context%3Aglobal+featureName+repo%3A%5Egithub%5C.com%2FUnleash%2Funleash-client-php%24+&patternType=standard&sm=1&groupBy=path

src/Repository/DefaultUnleashProxyRepository.php
```diff
@@ -0,0 +1,207 @@
...
+        $url = $this->addQuery($featureUrl, $this->contextToQueryString($context));
+
+        $request = $this->requestFactory->createRequest('GET', $url)
+            ->withHeader('Content-Type', 'application/json')
+            ->withHeader('Accept', 'application/json');
```
gastonfournier: I believe we'd only need the Accept header as it's a GET request, so there's no body content. But still, having an extra header doesn't hurt, so not a blocker

src/Repository/DefaultUnleashProxyRepository.php
```diff
@@ -0,0 +1,207 @@
...
+            $request = $request->withHeader($name, $value);
+        }
+
+        $apiKey = $this->configuration->getProxyKey();
+        if ($apiKey === null) {
+            // The only way we can get here is if this is manually constructed without the builder
+            // @codeCoverageIgnoreStart
+            throw new LogicException('No api proxy key was specified');
+            // @codeCoverageIgnoreEnd
+        }
```
gastonfournier: Details: maybe we should check this a bit earlier in the method, but it's an edge case, so not a big deal if you just ignore my comment

src/Repository/DefaultUnleashProxyRepository.php
```diff
@@ -0,0 +1,207 @@
...
+
+        $request = $request->withHeader('Authorization', $apiKey);
+
+        $response = $this->httpClient->sendRequest($request);
+        $body = json_decode($response->getBody(), true);
```
gastonfournier, line 63: Should we check: `$response->getStatusCode() >= 200 && $response->getStatusCode() < 300` before trying to parse the body as json? I'm wondering if we receive an error, the body most likely will be a json but we'll fail down in the validateResponse. But we can also just get a text response from a proxy layer on top of Unleash server (nginx or apache or any load balancer). I think we should check the response code before trying to decode

src/Repository/DefaultUnleashProxyRepository.php
```diff
@@ -0,0 +1,207 @@
...
+
+        return null;
+    }
+
+    private function getCacheKey(string $featureName, ?Context $context): string
+    {
+        if ($context === null) {
+            return $featureName;
+        }
+        $contextHash = md5($this->contextToQueryString($context));
```
gastonfournier: Maybe I'm far down into details, but shouldn't we use something like [xxh128](https://php.watch/versions/8.1/xxHash) instead of md5? I know the latter is fast but benchmarks are showing xxh more than 15 times faster: https://php.watch/articles/php-hash-benchmark 

Ofc, we can choose another one if we want to support older PHP versions as xxh is available from 8.1 and if we want to continue to support PHP 7.2 or newer we might have to choose something different like crc32c (which will also reduce from 128b to 32b)

src/Repository/DefaultUnleashProxyRepository.php
```diff
@@ -0,0 +1,207 @@
...
+
+        // @codeCoverageIgnoreStart
+        if ($urlParts === false) {
+            throw new \InvalidArgumentException('Invalid URL provided');
+        }
+        // @codeCoverageIgnoreEnd
+
+        if (!isset($urlParts['query'])) {
+            $urlParts['query'] = $query;
+        }
```
gastonfournier, line 133: What happens if `$urlParts['query']` is set? Are we not merging `$query` with `$urlParts['query']`? If this shouldn't happen maybe we can add a condition to the if above which is checking `if ($urlParts === false)`: `if ($urlParts === false || isset($urlParts['query'])`

gastonfournier (commented):
I think most of my comments are details, I'm more than happy to review again after you go over them

examples/09-cached-proxy.php
sighphyre: The little sleep here allows the metrics to get sent to the server. It's not a major need, it's nice to see in an example

src/Repository/DefaultUnleashProxyRepository.php
sighphyre: I don't _think_ this is a major issue. Hashing a small context object is going to be pretty cheap, and this is supported all the way back to the PHP versions we want to target. I'd rather have a few extra cpu cycles in exchange for wide support

src/Repository/DefaultUnleashProxyRepository.php
sighphyre: Then something super odd is going on and user has found a creative and interesting way to misuse this code. This url is built up by this code based on a root url that's passed in at startup. I'm pretty sure this should never be an issue

src/Repository/DefaultUnleashProxyRepository.php
sighphyre: Great feedback! Thanks, I'll fix

src/Repository/DefaultUnleashProxyRepository.php
sighphyre: Easy enough to change, good idea, I'll patch this

src/Repository/DefaultUnleashProxyRepository.php
sighphyre: Fair enough, may as well clean that up

src/Repository/DefaultUnleashProxyRepository.php
sighphyre: It is the first time, yea. I don't think we need to url encode this. You shouldn't ever be targeting a non url safe feature name because you can't get one into unleash itself. If you misspell it, this won't work anyway even with a nicely encoded name

src/DefaultUnleash.php
sighphyre: I'd rather leave the existing code paths alone to be honest. This is good feedback but I'd rather do this in a separate PR if we tackle it

examples/09-cached-proxy.php
RikudouSage: This one was correct, the `Psr16Cache` class wraps a PSR-6 cache and turns it into PSR-16 cache.

src/Repository/DefaultUnleashProxyRepository.php
RikudouSage: What about hash collisions? Md5 is pretty prone to those. I usually use `hash('sha512', $stringToHash)`.

src/Configuration/UnleashConfiguration.php
```diff
@@ -90,6 +91,18 @@ public function getTtl(): int
         return $this->ttl;
     }
 
+    public function getProxyKey(): ?string
+    {
+        return $this->proxyKey;
+    }
+
+    public function setProxyKey(string $proxyKey): self
```
RikudouSage: I think this should accept a nullable string as it's not a mandatory part of the configuration.

src/DTO/DefaultProxyFeature.php
```diff
@@ -0,0 +1,73 @@
...
+     */
+    public function __construct(array $response)
+    {
+        // This is validated elsewhere, this should only happen if a consumer
+        // tries to new this up manually and isn't interesting to tests
+
+        // @codeCoverageIgnoreStart
+        if (!isset($response['name'], $response['enabled'], $response['variant'], $response['impression_data'])) {
+            throw new \InvalidArgumentException('Invalid response structure');
+        }
```
RikudouSage: You may replace this with an assert which becomes a noop with production settings. But it's fine like this as well, really up to you.

src/DefaultProxyUnleash.php
```diff
@@ -0,0 +1,53 @@
...
+    ) {
+    }
+
+    public function register(): bool
+    {
+        //This is a no op, since registration is handled by the proxy/edge, this doesn't need coverage
+        // @codeCoverageIgnoreStart
+        return false;
+        // @codeCoverageIgnoreEnd
+    }
```
RikudouSage:
```suggestion
    /**
     * @codeCoverageIgnore
     */
    public function register(): bool
    {
        //This is a no op, since registration is handled by the proxy/edge, this doesn't need coverage
        return false;
    }
```

src/DefaultUnleash.php
```diff
@@ -115,14 +116,14 @@ public function isEnabled(string $featureName, ?Context $context = null, bool $d
         return false;
     }
 
-    public function getVariant(string $featureName, ?Context $context = null, ?Variant $fallbackVariant = null): Variant
+    public function getVariant(string $featureName, ?Context $context = null, ?ResolvedVariant $fallbackVariant = null): ResolvedVariant
```
RikudouSage: This is a BC break and requires a major version release. Can it extend the `Variant` interface? Or can we rename them so there's no BC break?

Overall I'm happy with a new major but it would take longer because there's a lot of cleanup I have to do for a new major release.

src/Repository/DefaultUnleashProxyRepository.php
```diff
@@ -0,0 +1,209 @@
...
+        $url = $this->addQuery($featureUrl, $this->contextToQueryString($context));
+
+        $request = $this->requestFactory->createRequest('GET', $url)
+            ->withHeader('Accept', 'application/json');
+
+        foreach ($this->configuration->getHeaders() as $name => $value) {
+            $request = $request->withHeader($name, $value);
+        }
+
+        $request = $request->withHeader('Authorization', $apiKey);
```
RikudouSage, line 56: Not sure we should add special support for api key, given that it's not supported in the main sdk. But we can also do it and raise an issue for adding the same support for api key in the main sdk as well.

src/Unleash.php
```diff
@@ -3,15 +3,15 @@
...
+use Unleash\Client\DTO\ResolvedVariant;
 
 interface Unleash
 {
     public const SDK_VERSION = '1.11.2';
 
     public function isEnabled(string $featureName, ?Context $context = null, bool $default = false): bool;
 
-    public function getVariant(string $featureName, ?Context $context = null, ?Variant $fallbackVariant = null): Variant;
+    public function getVariant(string $featureName, ?Context $context = null, ?ResolvedVariant $fallbackVariant = null): ResolvedVariant;
```
RikudouSage: A BC break, same as in previous comment.

src/UnleashBuilder.php
```diff
@@ -363,6 +366,12 @@ public function withVariantHandler(?VariantHandler $variantHandler): self
         return $this->with('variantHandler', $variantHandler);
     }
 
+    #[Pure]
+    public function withProxy(string $proxyKey): self
```
RikudouSage: This should IMO accept `?string`.

src/UnleashBuilder.php
```diff
@@ -525,14 +534,28 @@ public function build(): Unleash
...
-        return new DefaultUnleash(
-            $this->strategies,
-            $repository,
-            $registrationService,
-            $configuration,
-            $metricsHandler,
-            $variantHandler,
-        );
+        if ($this->proxyKey !== null) {
+            $configuration = $configuration->setProxyKey($this->proxyKey);
```
RikudouSage: Configuration is a stateful object, no need to reassign:

```suggestion
            $configuration->setProxyKey($this->proxyKey);
```

src/UnleashBuilder.php
```diff
@@ -525,14 +534,28 @@ public function build(): Unleash
...
+            $proxyRepository = new DefaultUnleashProxyRepository(
+                $configuration,
+                $httpClient,
+                $requestFactory
```
RikudouSage: Trailing comma for consistency:

```suggestion
                $requestFactory,
```

src/UnleashBuilder.php
```diff
@@ -525,14 +534,28 @@ public function build(): Unleash
...
+            );
+
+            return new DefaultProxyUnleash(
+                $proxyRepository,
+                $metricsHandler
```
RikudouSage:
```suggestion
                $metricsHandler,
```

src/UnleashBuilder.php
```diff
@@ -552,10 +575,12 @@ private function initializeServices(object $target, UnleashBuilderContainer $con
...
-                throw new CyclicDependencyException(sprintf(
-                    "A dependency '%s' is tagged as ConfigurationAware but that would cause a cyclic dependency as it needs to be part of Configuration",
-                    $target::class,
-                ));
+                throw new CyclicDependencyException(
+                    sprintf(
+                        "A dependency '%s' is tagged as ConfigurationAware but that would cause a cyclic dependency as it needs to be part of Configuration",
+                        $target::class,
+                    )
+                );
```
RikudouSage: Why this change?

src/UnleashBuilder.php
```diff
@@ -565,10 +590,12 @@ private function initializeServices(object $target, UnleashBuilderContainer $con
...
-                throw new CyclicDependencyException(sprintf(
-                    "A dependency '%s' is tagged as MetricsSenderAware but MetricsSender is not available for this type of dependency",
-                    $target::class,
-                ));
+                throw new CyclicDependencyException(
+                    sprintf(
+                        "A dependency '%s' is tagged as MetricsSenderAware but MetricsSender is not available for this type of dependency",
+                        $target::class,
+                    )
+                );
```
RikudouSage: Why this change?

tests/UnleashBuilderTest.php
```diff
@@ -283,11 +284,16 @@ public function testWithHeader()
...
+        self::assertNotSame(
+            $this->instance,
+            $this->instance->withRegistrationService(
+                new DefaultRegistrationService(
+                    $this->newHttpClient(),
+                    $this->newRequestFactory(),
+                    new UnleashConfiguration('', '', '')
+                )
+            )
+        );
```
RikudouSage: Why this change?

tests/UnleashBuilderTest.php
```diff
@@ -715,14 +721,20 @@ public function onNotFound(FeatureToggleNotFoundEvent $event): void
...
+                new IsIdentical($eventSubscriber2)
+            )
+        );
+        self::assertThat(
+            $subscribers[1],
+            self::logicalOr(
+                new IsIdentical($eventSubscriber1),
+                new IsIdentical($eventSubscriber2)
+            )
+        );
```
RikudouSage: Why this change?

RikudouSage (changes requested):
Overall it looks good, just leaving some comments.

examples/09-cached-proxy.php
sighphyre: I... I've rolled this one back, I'm a banana

src/Repository/DefaultUnleashProxyRepository.php
sighphyre: > What about hash collisions?
That's a decent argument. MD5 is often good enough but I'll happily take a better hash that's widely supported. I've swapped this out for sha512

src/Repository/DefaultUnleashProxyRepository.php
sighphyre: I would really like to go with this if it's okay. Having the key set explicitly is a lot more clear and now that we have two different key types in the mix, it's worth it to distinguish them. I'll open a ticket for adding the same thing for client keys

src/Configuration/UnleashConfiguration.php
sighphyre: Fair enough! Updated to be nullable

src/UnleashBuilder.php
sighphyre: Thanks!

src/UnleashBuilder.php
sighphyre: Also fair enough, also updated to be nullable

src/DTO/DefaultProxyFeature.php
sighphyre: That's a very nice feature and a good suggestion, I've swapped this out for an assert

src/DefaultUnleash.php
sighphyre: > Or can we rename them so there's no BC break?
So I tried that, it results in a lot of internal changes and make the whole PR ugly to read, I don't think either of us liked it much.

I'd really rather have this out sooner rather than later and we can absolutely make this change in a future PR, so I've made this rolled this back to the original Variant DTO and provided default parameters to the places in DefaultProxyUnleash that return this.

src/UnleashBuilder.php
sighphyre: I had thought this was fixer but apparently it was caused by something in my toolset in my IDE. I've rolled this back

src/UnleashBuilder.php
sighphyre: Same as above, rolled back

tests/UnleashBuilderTest.php
sighphyre: Same as above, rolled back

tests/UnleashBuilderTest.php
sighphyre: Same as above, rolled back

src/Unleash.php
sighphyre: Yeah, dealt with in the same way - rolled back for now, when we do a major bump, we can include this if we feel it's still important

RikudouSage:
Everything looks good to me now, let me know if you're done with all the changes and I'll be happy to merge.

sighphyre:
> Everything looks good to me now, let me know if you're done with all the changes and I'll be happy to merge.

Yes! I'm all happy, shall we merge and release this?

RikudouSage (approved):

PR #166, opened 2023-07-21
Create MetricsBucket interface to make MetricsSender usable externally

okhoshi:
# Description

As discussed on Slack, introduce an interface for `MetricsBucket` so `MetricsSender` interface doesn't depend on an internal class, letting external implementations of `MetricsHandler` free from using internal `MetricsBucket` (now named `DefaultMetricsBucket`).

## Type of change

# How Has This Been Tested?

- [x] Unit tests
- [x] Spec Tests

(I ran the full suite included in the repo)

# Checklist:


src/Metrics/DefaultMetricsBucket.php
```diff
@@ -0,0 +1,82 @@
...
+
+        return $this;
+    }
+
+    public function getStartDate(): DateTimeInterface
+    {
+        return $this->startDate;
+    }
+
+    public function setEndDate(?DateTimeInterface $endDate): DefaultMetricsBucket
```
RikudouSage: Can you return `self`? I noticed that it's a mistake carried over from my code, but anyway it would be better to make it consistent.

```suggestion
    public function setEndDate(?DateTimeInterface $endDate): self
```

src/Metrics/DefaultMetricsHandler.php
```diff
@@ -31,22 +31,22 @@ public function handleMetrics(Feature $feature, bool $successful, Variant $varia
         }
     }
 
-    private function getOrCreateBucket(): MetricsBucket
+    private function getOrCreateBucket(): DefaultMetricsBucket
```
RikudouSage, line 34: This should stay unchanged, meaning the return type hint should now point to the interface.

```suggestion
    private function getOrCreateBucket(): MetricsBucket
```

src/Metrics/DefaultMetricsHandler.php
```diff
@@ -31,22 +31,22 @@ public function handleMetrics(Feature $feature, bool $successful, Variant $varia
...
     {
         $cache = $this->configuration->getCache();
 
         $bucket = null;
         if ($cache->has(CacheKey::METRICS_BUCKET)) {
             $bucket = $cache->get(CacheKey::METRICS_BUCKET);
-            assert($bucket instanceof MetricsBucket || $bucket === null);
+            assert($bucket instanceof DefaultMetricsBucket || $bucket === null);
```
RikudouSage, line 41:
```suggestion
            assert($bucket instanceof MetricsBucket || $bucket === null);
```

src/Metrics/DefaultMetricsHandler.php
```diff
@@ -31,22 +31,22 @@ public function handleMetrics(Feature $feature, bool $successful, Variant $varia
...
         }
 
-        $bucket ??= new MetricsBucket(new DateTimeImmutable());
+        $bucket ??= new DefaultMetricsBucket(new DateTimeImmutable());
 
         return $bucket;
     }
 
-    private function shouldSend(MetricsBucket $bucket): bool
+    private function shouldSend(DefaultMetricsBucket $bucket): bool
```
RikudouSage, line 49:
```suggestion
    private function shouldSend(MetricsBucket $bucket): bool
```

src/Metrics/DefaultMetricsHandler.php
```diff
@@ -58,7 +58,7 @@ private function shouldSend(MetricsBucket $bucket): bool
         return $diff >= $this->configuration->getMetricsInterval();
     }
 
-    private function send(MetricsBucket $bucket): void
+    private function send(DefaultMetricsBucket $bucket): void
```
RikudouSage, line 61:
```suggestion
    private function send(MetricsBucket $bucket): void
```

src/Metrics/DefaultMetricsHandler.php
```diff
@@ -68,7 +68,7 @@ private function send(MetricsBucket $bucket): void
         }
     }
 
-    private function store(MetricsBucket $bucket): void
+    private function store(DefaultMetricsBucket $bucket): void
```
RikudouSage, line 71:
```suggestion
    private function store(MetricsBucket $bucket): void
```

src/Metrics/MetricsBucket.php
```diff
@@ -2,81 +2,7 @@
...
-use DateTimeInterface;
-use JetBrains\PhpStorm\ArrayShape;
-use JsonSerializable;
-use LogicException;
-
-/**
- * @internal
- */
-final class MetricsBucket implements JsonSerializable
+interface MetricsBucket
```
RikudouSage: I'd personally add the JsonSerializable interface here, or is there any reason why not to? Also this interface definitely needs some methods that will be used by the DefaultMetricsSender - it should work with any instance of `MetricsBucket`, meaning you need to declare all common methods here.

```suggestion
use JsonSerializable;

interface MetricsBucket extends JsonSerializable
```

RikudouSage (changes requested):
Thanks for the PR! Overall it seems it's going in the right direction, though the link between the interface and concrete implementations needs a bit of work.

src/Metrics/DefaultMetricsHandler.php
okhoshi: Then it would mean the `MetricsBucket` interface must expose an `addToggle` function, which takes a `MetricsBucketToggle` as an argument, which in turns need to be transformed in an interface... while the fact that the Bucket works with MetricsBucketToggle is an implementation detail IMHO.

It is also an implementation detail that the `DefaultMetricsHandler` works with `DefaultMetricsBucket`, and since that function is private, I don't think there's any benefit of returning the interface instead of the concrete class :)

src/Metrics/DefaultMetricsHandler.php
okhoshi: Same as my comment above

src/Metrics/MetricsBucket.php
okhoshi: The `JsonSerializable` interface is only needed in case you want to customise the serialisation output, it is not a prerequisite to use `json_encode` on an object.

As for the functions, the `DefaultMetricsSender` only needs to serialise the bucket with `json_encode`. The `MetricsBucket` is a simple DTO, is there really any function needed to be exposed ?

RikudouSage: Every JSON serializable object in the SDK uses the interface directly, this should be no exception.

As for the methods, yes, they are needed if you want to work with them in a universal way. If the interface is going to be public, it should definitely have getters, setters and a method to add a toggle to the bucket. Also a method that returns the data.

okhoshi: > Every JSON serializable object in the SDK uses the interface directly, this should be no exception.

Alright, I'll add for consistency then 👍

okhoshi:
@RikudouSage I added the interface for MetricsBucketToggle, does that look closer to what you had in mind ? :)

okhoshi:
@RikudouSage is there anything missing or that I should change on this PR ?

okhoshi:
Can we reopen this PR ?

PR #168, opened 2023-08-07
docs: some information on how to use the proxy sdk

sighphyre:
Adds some documentation for how to use the Proxy SDK

### DIFF
--- a/README.md
+++ b/README.md
@@ -292,6 +292,36 @@ $builder = UnleashBuilder::create()
     ]);
 ```
 
+
+## Proxy SDK
+
+By default the SDK uses the Server-side endpoints on the Unleash API. You can also use the Proxy SDK, which is a
+lightweight SDK that uses the Client-side endpoints on the Unleash API. The Proxy SDK give a substantial performance improvement when using a large set of feature toggles (10K+).
+
+To use the Proxy SDK, you need to call `withProxy($apiKey)` on the builder. The `$apiKey` needs to be a [frontend token](https://docs.getunleash.io/reference/api-tokens-and-client-keys#front-end-tokens). Note that `withProxy($apiKey)` is in lieu of setting the API key header.
+
+Example of using the builder to create a Proxy SDK instance:
+
+```php
+<?php
+$builder = UnleashBuilder::create()
+    ->withAppName('Some app name')
+    ->withAppUrl('https://some-app-url.com/api')
+    ->withInstanceId('Some instance id')
+    ->withProxy("some-proxy-key"); // <-- This is the only difference
+
+$unleash = $builder->build();
+
+$unleash.isEnabled("some-feature");
+```
+
+As of version 1.12, the Proxy SDK requires [Edge](https://docs.getunleash.io/reference/unleash-edge), so the `appUrl` needs to point to the Edge server.
+
+Not supported in the Proxy SDK:
+- Custom strategies
+- Registration (this is handled by Edge)
+
+
 ## Caching
 
 It would be slow to perform a http request every time you check if a feature is enabled, especially in popular


README.md
```diff
@@ -292,6 +292,36 @@ $builder = UnleashBuilder::create()
     ]);
 ```
 
+
+## Proxy SDK
+
+By default the SDK uses the Server-side endpoints on the Unleash API. You can also use the Proxy SDK, which is a
+lightweight SDK that uses the Client-side endpoints on the Unleash API. The Proxy SDK give a substantial performance improvement when using a large set of feature toggles (10K+).
```
gastonfournier, line 299: > By default the SDK uses the Server-side endpoints on the Unleash API. You can also use the Proxy SDK, which is a lightweight SDK that uses the Client-side endpoints on the Unleash API

I don't know if we ever explained what a server-side endpoint is compared to a client-side endpoint, but it might be confusing without the context. Maybe we can say that client-side endpoints delegate the evaluation of the toggle to the server while server-side does the evaluation?

gastonfournier (approved):
Looks good just a minor comment

README.md
```diff
@@ -292,6 +292,36 @@ $builder = UnleashBuilder::create()
     ]);
 ```
 
+
+## Proxy SDK
+
+By default the SDK uses the Server-side endpoints on the Unleash API. You can also use the Proxy SDK, which is a
+lightweight SDK that uses the Client-side endpoints on the Unleash API. The Proxy SDK give a substantial performance improvement when using a large set of feature toggles (10K+).
+
+To use the Proxy SDK, you need to call `withProxy($apiKey)` on the builder. The `$apiKey` needs to be a [frontend token](https://docs.getunleash.io/reference/api-tokens-and-client-keys#front-end-tokens). Note that `withProxy($apiKey)` is in lieu of setting the API key header.
```
gardleopard, line 301: sholdn't $apiKey be $frontendToken here then?

README.md
```diff
@@ -292,6 +292,36 @@ $builder = UnleashBuilder::create()
...
+    ->withAppUrl('https://some-app-url.com/api')
+    ->withInstanceId('Some instance id')
+    ->withProxy("some-proxy-key"); // <-- This is the only difference
+
+$unleash = $builder->build();
+
+$unleash.isEnabled("some-feature");
+```
+
+As of version 1.12, the Proxy SDK requires [Edge](https://docs.getunleash.io/reference/unleash-edge), so the `appUrl` needs to point to the Edge server.
```
gardleopard, line 318: Is it not possible to point this directly to the unleash frontend api?

gardleopard (changes requested):
I think this should work by pointing to the frontend api of an unleash instance.

README.md
RikudouSage: It's in the Unleash documentation, isn't it? This SDK can't be a replacement for all of documentation. Maybe adding a link, but that's it, IMO.

RikudouSage (approved):
Looks good to me.

README.md
sighphyre: I don't think so, no. This SDK (like most of our SDKs) calls this an API key (it does for the Server side token as well), I'm not all that keen to change the domain language here and especially not for users that are already using an API key

README.md
sighphyre: Not currently, no, I'd like to get the frontend api patched in the near future

README.md
sighphyre: Honestly, I don't think it's worth unpacking this more here. It's not relevant what the Client-side endpoints do for the purposes of the documentation, only that a separate API is being used. This is unpacked a lot all over the Unleash documentation

gastonfournier: Maybe linking to the documentation can help

gardleopard (approved):

PR #169, opened 2023-08-07
fix: make proxy client respect cache ttl

sighphyre:
This allows the Proxy Client to respect the cache ttl set in the configuration, otherwise the cache items lives for as long as the cache feels like allowing it to live


RikudouSage (approved):
Looks good to me.

PR #172, opened 2023-08-16
fix: return disabled variant when feature evaluated false

### LINKED ISSUE #171: Bug: when feature is evaluated false, it should return disabled variant

### Describe the bug

Currently, when feature is evaluated false, it is still returning a variant from feature, but should return disabled variant.
Also added client specification test for it https://github.com/Unleash/client-specification/pull/57

### To reproduce

1. Update client-specification to 4.2.3 and run tests, which will fail

### Sample code (optional)

_No response_

### Version

1.12.1

### Expected behavior

Feature should return disabled variant, when feature is evaluated false

### Logs (optional)

_No response_

I am working on the fix, just putting up issue for documentation

sjaanus:
Fixes https://github.com/Unleash/unleash-client-php/issues/171

1. Updated client spec to 4.2.3
2. Now getVariant will return disabled variant when feature is evaluated to false.

### DIFF
--- a/src/DefaultUnleash.php
+++ b/src/DefaultUnleash.php
@@ -26,13 +26,14 @@ final class DefaultUnleash implements Unleash
      * @param iterable<StrategyHandler> $strategyHandlers
      */
     public function __construct(
-        private readonly iterable $strategyHandlers,
-        private readonly UnleashRepository $repository,
-        private readonly RegistrationService $registrationService,
+        private readonly iterable             $strategyHandlers,
+        private readonly UnleashRepository    $repository,
+        private readonly RegistrationService  $registrationService,
         private readonly UnleashConfiguration $configuration,
-        private readonly MetricsHandler $metricsHandler,
-        private readonly VariantHandler $variantHandler,
-    ) {
+        private readonly MetricsHandler       $metricsHandler,
+        private readonly VariantHandler       $variantHandler,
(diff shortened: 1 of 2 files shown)


src/DefaultUnleash.php
```diff
@@ -39,16 +40,60 @@ public function __construct(
...
+        $feature = $this->findFeature($featureName);
+        return $this->isFeatureEnabled($feature, $default);
+    }
+
+    public function getVariant(string $featureName, ?Context $context = null, ?Variant $fallbackVariant = null): Variant
+    {
+        $fallbackVariant ??= $this->variantHandler->getDefaultVariant();
         $context ??= $this->configuration->getContextProvider()->getContext();
 
+        $feature = $this->findFeature($featureName);
```
kwasniew: same as lines 43 and 44? Can we call isEnabled?

src/DefaultUnleash.php
```diff
@@ -39,16 +40,60 @@ public function __construct(
 
     public function isEnabled(string $featureName, ?Context $context = null, bool $default = false): bool
```
kwasniew, line 41: why is context passed if we don't use it?

sighphyre (approved):
I think this looks good

sjaanus:
Codacy review seems not relevant.

PR #173, opened 2023-08-16
feat: add strategy variants support

sjaanus:
Add support for strategy variants.

https://docs.getunleash.io/reference/strategy-variants

To do it most efficiently, when feature is evaluated true, the underlying `isFeatureEnabled` method should return also the variant from the strategy returned true. Public contract remains the same.

### DIFF
--- a/src/DTO/DefaultStrategy.php
+++ b/src/DTO/DefaultStrategy.php
@@ -8,12 +8,14 @@ final class DefaultStrategy implements Strategy
      * @param array<string,string> $parameters
      * @param array<Constraint>    $constraints
      * @param array<Segment>       $segments
+     * @param array<Variant>       $variants
      */
     public function __construct(
         private readonly string $name,
         private readonly array $parameters = [],
         private readonly array $constraints = [],
         private readonly array $segments = [],
+        private readonly array $variants = [],
         private readonly bool $nonexistentSegments = false,
     ) {
     }
@@ -45,6 +47,14 @@ final class DefaultStrategy implements Strategy
     public function getSegments(): array
     {
         return $this->segments;
+    }
+
+    /**
+     * @return array<Variant>
+     */
+    public function getVariants(): array
+    {
+        return $this->variants;
     }
 
     public function hasNonexistentSegments(): bool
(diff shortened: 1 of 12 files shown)


sighphyre (changes requested):
This wants a bump to the passed client spec header in `DefaultUnleashRepository`, outside of that, LGTM

sjaanus:
> This wants a bump to the passed client spec header in `DefaultUnleashRepository`, outside of that, LGTM

Yes, good one.

sighphyre (approved):
LGTM!

src/DTO/DefaultStrategy.php
```diff
@@ -8,12 +8,14 @@ final class DefaultStrategy implements Strategy
...
      * @param array<Constraint>    $constraints
      * @param array<Segment>       $segments
+     * @param array<Variant>       $variants
      */
     public function __construct(
         private readonly string $name,
         private readonly array $parameters = [],
         private readonly array $constraints = [],
         private readonly array $segments = [],
+        private readonly array $variants = [],
```
RikudouSage, line 18: This is a backwards compatibility break, new parameters should be added as last.

src/DTO/DefaultStrategy.php
```diff
@@ -47,6 +49,14 @@ public function getSegments(): array
         return $this->segments;
     }
 
+    /**
+     * @return array<Variant>
+     */
+    public function getVariants(): array
+    {
+        return $this->variants;
+    }
```
RikudouSage, line 58: This is wrong, it's not the strategy that contains the variants, it's a feature that contains them.

RikudouSage (commented):
Well, could have gone better before merging.

src/DTO/DefaultStrategy.php
sjaanus: This is the new feature https://docs.getunleash.io/reference/strategy-variants. Now strategies also contain variants.

src/DTO/DefaultStrategy.php
sjaanus: Fixing it here https://github.com/Unleash/unleash-client-php/pull/174

PR #176, opened 2023-09-27
Dependent features

Tymek:
Edit: Moved this into draft so can iterate a little on it

# Description

Related to https://github.com/Unleash/unleash/issues/2255

Introducing concept of "dependency" on a feature toggle. Feature toggle, in order to be enabled an return a variant, can require the specified state of another feature toggle. There will be a `dependencies` property on each feature.

```
{
 ...
 "dependencies": {
   "feature": "string-featureId",
   "enabled: true, // boolean, state that is expected defaults to `true`
   "variants": ["variant1", "variant2"] // optional, variants allowed for feature to be enabled
 }
}
```

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# How Has This Been Tested?

- [ ] Unit tests
- [x] Spec Tests

# Checklist:


src/DefaultUnleash.php
```diff
@@ -59,21 +63,157 @@ public function isEnabled(string $featureName, ?Context $context = null, bool $d
...
+     */
+    public function isDependencySatisfied(
+        ?Dependency $dependency = null,
+        ?Feature $parentFeature = null,
+        ?Context $context = null,
+    ): bool {
+        if ($dependency === null) {
+            return true;
+        }
+        $context ??= $this->configuration->getContextProvider()->getContext();
```
kwasniew, line 92: you're passing $context in line 155 and now you're reassigning the same variable here. Is it intentional?

src/DefaultUnleash.php
```diff
@@ -59,21 +63,157 @@ public function isEnabled(string $featureName, ?Context $context = null, bool $d
             }
         }
 
-        return $this->isFeatureEnabled($feature, $context, $default)->isEnabled();
+        return $this->isFeatureEnabled($feature, $context, $dependencies, $default)->isEnabled();
```
kwasniew: why can't dependencies be taken from the feature and need to be passed as arguments

src/DefaultUnleash.php
RikudouSage: It's not a reassignment, it's the `null coalescing assign operator` (catchy name, innit?). It's basically shorthand for this:

```php
if (!isset($context)) {
  $context = $this->configuration->getContextProvider()->getContext();
}
```

Edit: Also, not sure how well versed in php you are, but `isset()`, contrary to what you might expect, doesn't check only for existence of variable but also whether it's `null`.

src/DefaultUnleash.php
```diff
@@ -59,21 +63,157 @@ public function isEnabled(string $featureName, ?Context $context = null, bool $d
...
+
+        if ($parentFeatureEnabled->isEnabled() && $dependency->getEnabled() === false) {
+            return false;
+        }
+        if (!$parentFeatureEnabled->isEnabled() && $dependency->getEnabled() !== false) {
+            return false;
+        }
+
+        $dependencyVariants = $dependency->getVariants();
+        if (!empty($dependencyVariants)) {
```
RikudouSage: Please don't ever use `empty()`, that does like a million things. Be specific, you probably want `count()` instead of `empty()`.

src/DefaultUnleash.php
kwasniew: @RikudouSage you're right, my bad :)

src/DefaultUnleash.php
```diff
@@ -43,7 +45,9 @@ public function __construct(
     public function isEnabled(string $featureName, ?Context $context = null, bool $default = false): bool
     {
         $context ??= $this->configuration->getContextProvider()->getContext();
-        $feature = $this->findFeature($featureName, $context);
+        $featureAndDependencies = $this->findFeatureAndDependencies($featureName, $context);
```
kwasniew: can findFeature return feature with dependencies same way as our API returns feature with dependencies?

src/DefaultUnleash.php
```diff
@@ -152,6 +266,18 @@ private function isFeatureEnabled(?Feature $feature, Context $context, bool $def
             return new DefaultFeatureEnabledResult();
         }
 
+        $dependencies = $feature->getDependencies();
+        if (!empty($dependencies)) {
+            foreach ($dependencies as $dependency) {
+                $parentFeature = $parentFeatures[$dependency->getFeature()] ?? null;
+                if (!$this->isDependencySatisfied($dependency, $parentFeature, $context)) {
+                    $this->metricsHandler->handleMetrics($feature, false);
```
kwasniew, line 241: this reminds me to add a test that verifies that metrics are not counted for the parent features

src/DefaultUnleash.php
```diff
@@ -59,21 +63,157 @@ public function isEnabled(string $featureName, ?Context $context = null, bool $d
...
+        if ($feature === null) {
+            $event = new FeatureToggleNotFoundEvent($context, $featureName);
+            $this->configuration->getEventDispatcherOrNull()?->dispatch(
+                $event,
+                UnleashEvents::FEATURE_TOGGLE_NOT_FOUND,
+            );
+
+            return [
+                'feature' => null,
+                'dependencies' => [],
```
kwasniew: actualDependencies

Tymek:
Can we somehow solve it in another way? Right now the issue is that it's not possible to get feature toggles from repository multiple times in multiple places in main class. This should be doable by holding in-memory copy of the response or state of the `Repository`, since it's already a property on main class. Am I missing something?
@sighphyre @RikudouSage

RikudouSage:
@Tymek I'm not entirely sure what this feature is about (no one told me), but it seems that features now can have a dependency on another feature. If any parent feature evaluates to false, all child features are also false. Is that a correct assumption?

In that case, why not construct it recursively? The DTOs pretty much should be the JSON response but as PHP DTOs. 

There's also the method `findFeature()`, why can't you use that to fetch the toggles multiple times?

If you describe the issue in more detail, I might be able to assist further.

src/DefaultUnleash.php
```diff
@@ -59,21 +63,157 @@ public function isEnabled(string $featureName, ?Context $context = null, bool $d
...
+     *
+     * @param string  $featureName name of the feature to find
+     * @param Context $context     the context to use
+     *
+     * @return null|array{
+     *   feature: Feature|null,
+     *   dependencies: array<string, Feature>,
+     * }
+     */
+    public function findFeatureAndDependencies(string $featureName, ?Context $context): ?array
```
sighphyre: Does this have to be public?

Also, I'm not sure this is the right layer to be doing this in. Not hitting the cache twice? Yes! Doing this... less yes.

Can we not materialise this from cache in a more complete state rather? I want to say the `parseFeatures` method in the repository is a good target for this. That can return a DTO that looks something like (in very pseudocode)

DefaultFeature {
  name: String,
  parent?: DefaultFeature
}

sighphyre: Also, I don't think this method should take context. It's strange. It doesn't affect the outcome in any way outside of meta data in the events. Especially for a function marked public that feels off

src/DefaultUnleash.php
sighphyre: Me: *Arrives on this pull request*
Me: *Reads this comment*
Me: Can `empty()` really be that bad?
Me: *Reads for 5 minutes*
Me: Please don't ever use `empty()`, that does like a million things

src/DefaultUnleash.php
sighphyre: I think we're thinking the same thing here, I have a similar comment on materializing this from cache in the correct DTO format. I _think_ we're saying the same thing. If we are, I definitely agree

src/Repository/DefaultUnleashRepository.php
```diff
@@ -83,9 +90,9 @@ public function findFeature(string $featureName): ?Feature
      * @throws ClientExceptionInterface
      * @throws JsonException
      *
-     * @return iterable<Feature>
+     * @return array<Feature>
```
sighphyre: I see we're shifted from an iterable to an array in a few places in this PR but I can't see why we would need to. What's happening here?

sighphyre (commented):
I don't believe it's enough to rely on the client specs here. This is feature that can go wrong in a lot of subtle ways, even though it's simple on the surface. The PHP implementation is always going to have details that most of the other SDK implementations don't - you know the details better than anyone, cover them with tests

If not to make me happy, then to make the quality gates on this repo happy

sighphyre (commented):
Can we also put a proper description on this PR? This is a significant feature and it's helpful for others to understand what it does, why it does it and how it works. Especially for reviewers who are seeing this feature for the first time

sighphyre:
> In that case, why not construct it recursively? The DTOs pretty much should be the JSON response but as PHP DTOs.

I think we can and should do that!

> There's also the method `findFeature()`, why can't you use that to fetch the toggles multiple times?

`findFeature()` will load and parse everything from cache every time, it's expensive(ish) with large feature sets. I'd rather avoid it if we can. If it's hard to do, I'm happy to go with better, more understandable code over fast code. But I think just constructing the DTOs in the repo layer recursively would give us nice code that's also cheap

RikudouSage:
Replaced with #187

PR #179, opened 2023-10-30
fix: use a fixed hash seed for variant hashing

sighphyre:
This forces the variant distribution to use a seeded hash, rather than falling back to the default of 0.

This fixes an issue in the variant distribution when gradual rollout is at play, since that narrows down the potential buckets before variants kick in. Previously, we'd see very non random distributions of variants in those cases. This change allows for a more statistically sane distribution

This PR hasn't updated the client-spec, since this requires dependent features to get merged first in order for the specs to pass. Manually validated by removing the dependent features test case and running the rest of the tests. Because of this, this PR won't pass until that feature gets merged or we backport a version for the client spec

### DIFF
--- a/src/Stickiness/StickinessCalculator.php
+++ b/src/Stickiness/StickinessCalculator.php
@@ -4,5 +4,5 @@ namespace Unleash\Client\Stickiness;
 
 interface StickinessCalculator
 {
-    public function calculate(string $id, string $groupId, int $normalizer = 100): int;
+    public function calculate(string $id, string $groupId, int $normalizer = 100, int $seed = 0): int;
 }
--- a/src/Variant/DefaultVariantHandler.php
+++ b/src/Variant/DefaultVariantHandler.php
@@ -11,6 +11,8 @@ use Unleash\Client\Stickiness\StickinessCalculator;
 
 final class DefaultVariantHandler implements VariantHandler
 {
+    const VARIANT_HASH_SEED = 86028157;
+
     public function __construct(
         private readonly StickinessCalculator $stickinessCalculator,
     ) {
@@ -94,7 +96,7 @@ final class DefaultVariantHandler implements VariantHandler
         int $totalWeight
     ): int {
         $stickiness = $variants[0]->getStickiness();
-        if ($stickiness !== Stickiness::DEFAULT) {
+        if ($stickiness !== Stickiness::DEFAULT ) {
             $seed = $context->findContextValue($stickiness) ?? $this->randomString();
         } else {
             $seed = $context->getCurrentUserId()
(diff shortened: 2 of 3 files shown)


src/Stickiness/StickinessCalculator.php
```diff
@@ -4,5 +4,5 @@
 
 interface StickinessCalculator
 {
-    public function calculate(string $id, string $groupId, int $normalizer = 100): int;
+    public function calculate(string $id, string $groupId, int $normalizer = 100, int $seed = 0): int;
```
RikudouSage, line 7: This would need a major version release as it's a BC break.

src/Variant/DefaultVariantHandler.php
```diff
@@ -94,7 +96,7 @@ private function calculateStickiness(
         int $totalWeight
     ): int {
         $stickiness = $variants[0]->getStickiness();
-        if ($stickiness !== Stickiness::DEFAULT) {
+        if ($stickiness !== Stickiness::DEFAULT ) {
```
RikudouSage: Probably just a forgotten space.

src/Variant/DefaultVariantHandler.php
```diff
@@ -11,6 +11,8 @@
 
 final class DefaultVariantHandler implements VariantHandler
 {
+    const VARIANT_HASH_SEED = 86028157;
```
RikudouSage: Where is this value from? Also, please add visibility (`private`, `protected`, `public`)

src/Stickiness/StickinessCalculator.php
sighphyre: Yes, yes it will. We think that's correct anyway since this will trigger a new distribution of variants and that's likely to be a significant change for anyone using the SDK

So probably time to discuss that new major version then

src/Variant/DefaultVariantHandler.php
sighphyre: Ugh, VSCode "helping" me again. Thanks, good spot

src/Variant/DefaultVariantHandler.php
sighphyre: Marked it as private

This is just an arbitrary but large number used for the seed. We have the same bug in all of our SDKs so we're standardising on this

I'd love a way to communicate this out with the client spec but that isn't really possible with the client specs design right now

src/Stickiness/StickinessCalculator.php
RikudouSage: Definitely, there are a few changes racked up over the SDK's life that are just waiting for a new major release.

RikudouSage (approved):

PR #183, opened 2023-11-08
Do not store invalid response

vbartusevicius:
# Description

When the `max_duration` option is used in the PSR18 client, it may result in a partial/corrupted body that is stored without any checks for its validity. Then in `DefaultUnleashRepository:parseFeatures`, the JsonException is thrown and is bubbled up for the end-user. Additionally, the Stale cache is polluted.

More on the error: https://github.com/symfony/symfony/issues/52490

Fixes # (issue)

## Type of change

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# How Has This Been Tested?

- [x] Unit tests
- [ ] Spec Tests

# Checklist:

### DIFF
--- a/src/Repository/DefaultUnleashRepository.php
+++ b/src/Repository/DefaultUnleashRepository.php
@@ -180,6 +180,7 @@ final readonly class DefaultUnleashRepository implements UnleashRepository
 
     /**
      * @throws JsonException
+     * @param array $body
      *
      * @return array<Feature>
      */
--- a/src/Repository/DefaultUnleashRepository.php
+++ b/src/Repository/DefaultUnleashRepository.php
@@ -55,6 +55,24 @@ use Unleash\Client\Exception\InvalidValueException;
  *       type:string,
  *       value: string,
  *   }
+ * @phpstan-type StrategyArray array{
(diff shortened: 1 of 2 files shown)


src/Repository/DefaultUnleashRepository.php
```diff
@@ -109,6 +109,7 @@ public function getFeatures(): iterable
                     $response = $this->httpClient->sendRequest($request);
                     if ($response->getStatusCode() === 200) {
                         $data = (string) $response->getBody();
+                        json_decode($data, true, 512, JSON_THROW_ON_ERROR);
```
RikudouSage: Sadly it needs to work on 7.2 as well, which doesn't support the `JSON_THROW_ON_ERROR` flag. This needs to be checked manually using [json_last_error](https://www.php.net/manual/en/function.json-last-error.php).

I also don't like the idea that decoding json would be done twice, so it would be best to change the `parseFeatures()` method to accept already `json_decode`d array instead of a string.

RikudouSage (approved):
Looks good, thank you for your contribution!

PR #185, opened 2023-11-14
fix: proxy sdk now sends metrics to the correct endpoint

sighphyre:
# Description

This patches a bug in the metrics sending code for the case where we set the SDK into Proxy mode. 

Previously, the metrics code would send metrics to the same endpoint regardless of the mode that it was operating in. However, Edge does not accept metrics for frontend clients on the same URL as backend clients.

This patches the code so that the URL to send metrics to is determined at runtime based on whether or not the SDK is in Proxy mode or not

## Type of change

# How Has This Been Tested?

- [x] Unit tests
- [ ] Spec Tests

# Checklist:


RikudouSage (approved):
Looks good!

PR #187, opened 2023-11-24
Feat: Dependent features

RikudouSage:
# Description

Adds support for dependent features.

Replaces #176 

## Type of change

# How Has This Been Tested?

- [x] Unit tests
- [x] Spec Tests

# Checklist:

### DIFF
--- a/src/DefaultUnleash.php
+++ b/src/DefaultUnleash.php
@@ -7,6 +7,7 @@ use Unleash\Client\Configuration\Context;
 use Unleash\Client\Configuration\UnleashConfiguration;
 use Unleash\Client\DTO\DefaultFeatureEnabledResult;
 use Unleash\Client\DTO\Feature;
+use Unleash\Client\DTO\FeatureDependency;
 use Unleash\Client\DTO\FeatureEnabledResult;
 use Unleash\Client\DTO\Strategy;
 use Unleash\Client\DTO\Variant;
@@ -152,6 +153,24 @@ final readonly class DefaultUnleash implements Unleash
             return new DefaultFeatureEnabledResult();
         }
 
+        $dependencies = method_exists($feature, 'getDependencies')
+            ? $feature->getDependencies()
+            : [];
+
+        foreach ($dependencies as $dependency) {
+            if ($this->isDependencyEnabled($dependency, $context, $default) !== $dependency->getExpectedState()) {
+                $event = new FeatureToggleDisabledEvent($feature, $context);
+                $this->configuration->getEventDispatcher()->dispatch(
+                    $event,
+                    UnleashEvents::FEATURE_TOGGLE_DISABLED,
+                );
+
+                $this->metricsHandler->handleMetrics($feature, false);
+
+                return new DefaultFeatureEnabledResult();
+            }
+        }
+
         $strategies = $feature->getStrategies();
         if (!is_countable($strategies)) {
             $strategies = iterator_to_array($strategies);
@@ -204,5 +223,36 @@ final readonly class DefaultUnleash implements Unleash
         }
 
         return $handlers;
+    }
+
+    private function isDependencyEnabled(FeatureDependency $dependency, Context $context, bool $default): bool
+    {
+        if (!$dependency->isResolved()) {
+            return false;
+        }
+
+        $enabled = $this->isFeatureEnabled($dependency->getFeature(), $context, $default);
+        if (!$enabled->isEnabled()) {
+            return false;
+        }
+
+        assert($dependency->getFeature() !== null);
+
+        if (
+            method_exists($dependency->getFeature(), 'getDependencies')
+            && count($dependency->getFeature()->getDependencies())
+        ) {
+            return false;
+        }
+
+        if ($dependency->getRequiredVariants() === null || !count($dependency->getRequiredVariants())) {
+            return true;
+        }
+
+        $variant = $this->getVariant($dependency->getFeature()->getName(), $context);
+
+        $requiredVariants = array_map(fn (Variant $variant) => $variant->getName(), $dependency->getRequiredVariants());
+
+        return in_array($variant->getName(), $requiredVariants, true);
     }
 }
--- a/src/DefaultUnleash.php
+++ b/src/DefaultUnleash.php
@@ -158,7 +158,7 @@ final readonly class DefaultUnleash implements Unleash
             : [];
 
         foreach ($dependencies as $dependency) {
-            if ($this->isDependencyEnabled($dependency, $context, $default) !== $dependency->getExpectedState()) {
+            if ($this->isParentDependencySatisfied($dependency, $context, $default) !== $dependency->getExpectedState()) {
                 $event = new FeatureToggleDisabledEvent($feature, $context);
                 $this->configuration->getEventDispatcher()->dispatch(
                     $event,
@@ -225,7 +225,7 @@ final readonly class DefaultUnleash implements Unleash
         return $handlers;
     }
 
-    private function isDependencyEnabled(FeatureDependency $dependency, Context $context, bool $default): bool
+    private function isParentDependencySatisfied(FeatureDependency $dependency, Context $context, bool $default): bool
     {
         if (!$dependency->isResolved()) {
             return false;
--- a/src/Repository/DefaultUnleashRepository.php
+++ b/src/Repository/DefaultUnleashRepository.php
@@ -13,12 +13,17 @@ use Unleash\Client\Configuration\UnleashConfiguration;
 use Unleash\Client\DTO\Constraint;
 use Unleash\Client\DTO\DefaultConstraint;
 use Unleash\Client\DTO\DefaultFeature;
+use Unleash\Client\DTO\DefaultFeatureDependency;
 use Unleash\Client\DTO\DefaultSegment;
 use Unleash\Client\DTO\DefaultStrategy;
 use Unleash\Client\DTO\DefaultVariant;
 use Unleash\Client\DTO\DefaultVariantOverride;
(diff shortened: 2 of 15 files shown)


tests/DefaultUnleashTest.php
```diff
@@ -813,6 +826,182 @@ public function testAutoRegistrationOnInvalidResponse()
         );
     }
 
+    public function testDependenciesUnresolvedFeatureVariants()
```
RikudouSage, line 829: This test might be worth adding to client specs.

tests/DefaultUnleashTest.php
```diff
@@ -813,6 +826,182 @@ public function testAutoRegistrationOnInvalidResponse()
...
+                    ],
+                ],
+            ],
+        ];
+        $this->pushResponse($data);
+        $instance = $this->getCachedInstance();
+        self::assertTrue($instance->isEnabled('child'));
+    }
+
+    public function testDependenciesDirectFeatureVariants()
```
RikudouSage, line 870: This test might be worth adding to client specs.

tests/DefaultUnleashTest.php
```diff
@@ -813,6 +826,182 @@ public function testAutoRegistrationOnInvalidResponse()
...
+                    ],
+                ],
+            ],
+        ];
+        $this->pushResponse($data);
+        $instance = $this->getCachedInstance();
+        self::assertTrue($instance->isEnabled('child'));
+    }
+
+    public function testDependenciesMixedResolvedUnresolved()
```
RikudouSage, line 911: This test might be worth adding to client specs.

src/DefaultUnleash.php
```diff
@@ -205,4 +224,35 @@ private function findStrategyHandlers(Strategy $strategy): array
 
         return $handlers;
     }
+
+    private function isDependencyEnabled(FeatureDependency $dependency, Context $context, bool $default): bool
```
kwasniew: can we adjust domain vocabulary to `isParentDependencySatisfied`? The reason for that is that we want to emphasize that this is a parent dependency not a child dependency. Also dependency doesn't have to be enabled - the spec also defines disabled parent as an option.

RikudouSage: Done.

src/Repository/DefaultUnleashRepository.php
```diff
@@ -239,19 +247,161 @@ private function parseFeatures(array $body): array
...
+        $this->handleUnresolvedDependencies($unresolved, $features);
+
         return $features;
     }
 
+    /**
+     * @param array<Feature>                $unresolved
+     * @param array<string, DefaultFeature> $features
+     */
+    private function handleUnresolvedDependencies(array &$unresolved, array &$features): void
```
kwasniew: This part seems to be very different from all the other Unleash SDKs I've seen. Do we have any README/design decisions/ADRs in this repo that would explain why we're doing this part?

RikudouSage: Not really, I haven't looked at the other SDKs and this is just how I did this.

tests/DefaultUnleashTest.php
kwasniew: I though "Child is enabled if its parent matches a specific variant." matches this case. Can you elaborate what exactly are we missing in the client spec? We'd like the spec to cover all corner cases so all the findings that you make are greatly appreciated.

RikudouSage: If I'm not mistaken, the spec tests only contain variants in strategies, not variants in features directly.

tests/DefaultUnleashTest.php
kwasniew: Can you confirm if "Child is enabled if its parent matches any of the specified variants." is covering this case?

tests/DefaultUnleashTest.php
kwasniew: That's correct since variants and strategy variants have separate spec that should prove that they behave the same. In the long term we'd like to migrate towards strategy variants and slowly sunset feature environment variants.

src/Repository/DefaultUnleashRepository.php
```diff
@@ -239,19 +247,161 @@ private function parseFeatures(array $body): array
...
+    {
+        if (count($unresolved)) {
```
kwasniew: What do you think about extracting some smaller methods explaining the steps of the algorithm?

Just for inspiration chatgpt session suggested sth like this so take it with a grain of salt :) 
```php
    private function handleUnresolvedDependencies(array &$unresolved, array &$features): void
    {
        if (empty($unresolved)) {
            return;
        }

        do {
            $currentUnresolvedCount = count($unresolved);
            foreach ($unresolved as $index => $unresolvedItem) {
                $dependencies = $this->resolveFeatureDependencies($unresolvedItem, $features);
                $this->updateFeatureWithResolvedDependencies($unresolvedItem, $dependencies, $features);
                unset($unresolved[$index]);
            }
        } while (count($unresolved) < $currentUnresolvedCount);
    }

    private function resolveFeatureDependencies($unresolvedItem, array &$features): array
    {
        $dependencies = [];
        foreach ($unresolvedItem->getDependencies() as $dependency) {
            if (!$dependency instanceof UnresolvedFeatureDependency) {
                continue;
            }

            $feature = $this->resolveDependencyFeature($dependency, $features);
            $requiredVariants = $this->resolveDependencyVariants($dependency, $feature);

            $dependencies[] = $this->createFeatureDependency($feature, $dependency, $requiredVariants);
        }
        return $dependencies;
    }

    private function resolveDependencyFeature($dependency, array &$features)
    {
        $feature = $dependency->getFeature();
        if ($feature instanceof UnresolvedFeature) {
            return $features[$feature->getName()] ?? null;
        }
        return $feature;
    }

    private function resolveDependencyVariants($dependency, $feature): array
    {
        $requiredVariants = [];
        foreach ($dependency->getRequiredVariants() ?? [] as $requiredVariant) {
            if (!$requiredVariant instanceof UnresolvedVariant) {
                $requiredVariants[] = $requiredVariant;
                continue;
            }
            $resolvedVariant = $this->findMatchingVariant($feature, $requiredVariant);
            $requiredVariants[] = $resolvedVariant ?: $requiredVariant;
        }
        return $requiredVariants;
    }

    private function findMatchingVariant($feature, $requiredVariant)
    {
        foreach ($feature->getVariants() as $variant) {
            if ($variant->getName() === $requiredVariant->getName()) {
                return $variant;
            }
        }
        return null;
    }

    private function createFeatureDependency($feature, $dependency, array $requiredVariants): DefaultFeatureDependency
    {
        return new DefaultFeatureDependency(
            feature: $feature,
            expectedState: $dependency->getExpectedState(),
            requiredVariants: $requiredVariants,
        );
    }

    private function updateFeatureWithResolvedDependencies($unresolvedItem, array $dependencies, array &$features): void
    {
        $features[$unresolvedItem->getName()] = new DefaultFeature(
            name: $unresolvedItem->getName(),
            enabled: $unresolvedItem->isEnabled(),
            strategies: $unresolvedItem->getStrategies(),
            variants: $unresolvedItem->getVariants(),
            impressionData: $unresolvedItem->hasImpressionData(),
            dependencies: $dependencies,
        );
    }
```

RikudouSage: Sure, I can split it into smaller methods.

src/Repository/DefaultUnleashRepository.php
```diff
@@ -239,19 +247,161 @@ private function parseFeatures(array $body): array
...
+        }
+    }
+
+    /**
+     * @param array<array{feature: string, enabled?: bool, variants?: array<string>}> $dependencies
+     * @param array<string, Feature>                                                  $features
+     *
+     * @return array<FeatureDependency>
+     */
+    private function parseDependencies(array $dependencies, array $features, ?bool &$hasUnresolvedDependencies = null): array
```
kwasniew, line 358: and same here, personally I'd do extract method refactoring to explain the steps of the algorithm and reduce the indentation level/whitespace on the left hand side of the code.

kwasniew (commented):
I'll do final review on Monday after the extract methods refactoring.

src/Repository/DefaultUnleashRepository.php
RikudouSage: I've instead rewritten the code and added comments, is that enough? Some parts of that code were from a previous version where I assumed infinite nesting of dependencies, I've removed them now.

src/Repository/DefaultUnleashRepository.php
RikudouSage: I tried to split it up a bit, hopefully it helps.

src/Repository/DefaultUnleashRepository.php
```diff
@@ -239,19 +247,196 @@ private function parseFeatures(array $body): array
...
         }
 
+        // $features are passed by reference so that the method can modify them
+        $this->handleUnresolvedDependencies($unresolved, $features);
+
         return $features;
     }
 
+    /**
+     * All unresolved dependencies are solved in this method (if possible).
```
kwasniew, line 274: This is fine for now. Every SDK will be much more complex with multi-level deps. However in our initial research we haven't found a need for this extra complexity.

kwasniew (approved):

src/Repository/DefaultUnleashRepository.php
RikudouSage: Yeah, it's just the for future me or some other future implementor in case it's needed.

kwasniew:
@RikudouSage I'll update our docs (https://docs.getunleash.io/reference/dependent-features#client-sdk-support) after you publish a new version

PR #188, opened 2023-11-26
Fix: Add a check for usage of the empty() construct

RikudouSage:
# Description

I've noticed some code which uses the `empty()` construct. The construct does way too many things and should not be used in any modern code. Additionally it looks like a function but isn't, which might cause problems in a few cases.

This PR adds a PHPStan rule that checks for usage of `empty()` and offers a hint if the type of variable is known.

# Checklist:


RikudouSage:
You can check what the output looks like here: https://github.com/Unleash/unleash-client-php/actions/runs/6997953019/job/19035484861?pr=188

sighphyre (approved):
Oh wow, that's pretty cool. Yes please!

PR #190, opened 2023-12-01
Update README.md with transpiled release info for PHP versions

markunl:
# Description

Updates README.md with transpiled release info guidance for users to match SDK releases to their version of PHP.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

Previewed readme changes for formatting.

# Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,11 @@ If you're migrating from 1.x to 2.x, you can read the [migration guide](doc/migr
 
 `composer require unleash/client`
 
-Requires PHP 7.2 or newer.
+Requires PHP 7.2 or newer.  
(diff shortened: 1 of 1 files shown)


RikudouSage:
Is that necessary? If you run it without any version, the biggest version available is installed - if you're on PHP 7.4, the version for 7.4 will be installed. Though it might be a problem if production uses different version than dev machine.

markunl:
> Is that necessary? If you run it without any version, the biggest version available is installed - if you're on PHP 7.4, the version for 7.4 will be installed. Though it might be a problem if production uses different version than dev machine.

Makes sense, we can drop this. Main motivation behind the update was a user asking for help and not being clear on which version of the SDK they should be using with their PHP version. But if this is handled automatically during install then I agree no update needed.

PR #191, opened 2023-12-19
fix: If withProxy(apiKey) is set the client does not set custom headers

chriswk:
Since our README explicitly states that using withProxy($apiKey) is in lieu of using custom headers for the Authorization header, our SDK should make sure to check if the proxyApiKey is set, and if so, use it to set the Authorization header also for client registration and client metrics.

# Description

We had clients complaining that when they used the PHP client in proxy mode (using the withProxy method), they struggled with 403 responses when trying to register/post metrics. After diving into the code I found that we were not using the key as an authorization header for any other request but the request to /api/frontend.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [x] Unit tests
- [ ] Spec Tests

# Checklist:

### DIFF
--- a/src/Client/DefaultRegistrationService.php
+++ b/src/Client/DefaultRegistrationService.php
@@ -57,7 +57,6 @@ final class DefaultRegistrationService implements RegistrationService
                 'started' => (new DateTimeImmutable())->format('c'),
                 'interval' => $this->configuration->getMetricsInterval(),
             ], JSON_THROW_ON_ERROR)));
-        
         foreach ($this->configuration->getHeaders() as $name => $value) {
             $request = $request->withHeader($name, $value);
         }
--- a/src/Client/DefaultRegistrationService.php
+++ b/src/Client/DefaultRegistrationService.php
@@ -57,6 +57,10 @@ final class DefaultRegistrationService implements RegistrationService
                 'started' => (new DateTimeImmutable())->format('c'),
                 'interval' => $this->configuration->getMetricsInterval(),
             ], JSON_THROW_ON_ERROR)));
+        
+        if ($this->configuration->getProxyKey() !== null) {
+            $request = $request->withHeader('Authorization', $this->configuration->getProxyKey());
+        }
         foreach ($this->configuration->getHeaders() as $name => $value) {
             $request = $request->withHeader($name, $value);
         }
--- a/src/Client/DefaultRegistrationService.php
+++ b/src/Client/DefaultRegistrationService.php
@@ -58,9 +58,6 @@ final class DefaultRegistrationService implements RegistrationService
                 'interval' => $this->configuration->getMetricsInterval(),
             ], JSON_THROW_ON_ERROR)));
         
-        if ($this->configuration->getProxyKey() !== null) {
-            $request = $request->withHeader('Authorization', $this->configuration->getProxyKey());
-        }
         foreach ($this->configuration->getHeaders() as $name => $value) {
             $request = $request->withHeader($name, $value);
         }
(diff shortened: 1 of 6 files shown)


src/Metrics/DefaultMetricsSender.php
```diff
@@ -31,12 +31,15 @@ public function sendMetrics(MetricsBucket $bucket): void
...
+        if ($this->configuration->getProxyKey() !== null) {
+            $request = $request->withHeader('Authorization', $this->configuration->getProxyKey());
+        }
         foreach ($this->configuration->getHeaders() as $name => $value) {
             $request = $request->withHeader($name, $value);
         }
 
         try {
-            $this->httpClient->sendRequest($request);
+            $response = $this->httpClient->sendRequest($request);
```
RikudouSage: Why the variable assignment?

src/Client/DefaultRegistrationService.php
```diff
@@ -57,6 +57,10 @@ public function register(iterable $strategyHandlers): bool
                 'started' => (new DateTimeImmutable())->format('c'),
                 'interval' => $this->configuration->getMetricsInterval(),
             ], JSON_THROW_ON_ERROR)));
+        
+        if ($this->configuration->getProxyKey() !== null) {
+            $request = $request->withHeader('Authorization', $this->configuration->getProxyKey());
+        }
```
RikudouSage: Well, this is rather unfortunate solution, but sadly I don't think we have any other quick fix. The proxy key should be directly tied to the headers in configuration.

Can you add a comment with `TODO: refactor proxy key into headers` or something similar, so that I'll pick it up next time I'm doing clean-up?

src/Metrics/DefaultMetricsSender.php
```diff
@@ -31,12 +31,15 @@ public function sendMetrics(MetricsBucket $bucket): void
                 'instanceId' => $this->configuration->getInstanceId(),
                 'bucket' => $bucket->jsonSerialize(),
             ], JSON_THROW_ON_ERROR)));
+        if ($this->configuration->getProxyKey() !== null) {
+            $request = $request->withHeader('Authorization', $this->configuration->getProxyKey());
+        }
```
RikudouSage: Add the same todo comment here, please.

RikudouSage (changes requested):

src/Metrics/DefaultMetricsSender.php
chriswk: Ah, thanks for the spot, because, I needed to see the status code, it wasn't supposed to be part of the PR, reverting :)

src/Client/DefaultRegistrationService.php
chriswk: I can have another stab before converting this to a proper PR.

chriswk: Totally agree that this should be moved to the headers list when we build the configuration.

src/Metrics/DefaultMetricsSender.php
chriswk: I'm going to have one more stab at the builder before converting this to a proper PR. Thanks for the feedback :+1:

chriswk:
Hey @RikudouSage - could I get a stop or a go-ahead here, I added the api key to the headers in the builder's build method instead, which meant I could remove the unfortunate if checks in our request handling.

RikudouSage (approved):
Looks good to me! Do you want to do any further changes or shall I merge?

chriswk:
Please merge, sorry, I see I owed you this answer.

PR #198, opened 2024-02-07
Feat: Compact metrics

### LINKED ISSUE #197: Feat: Optimize Metrics cache structure to be resource-friendly on bigger usages

### Describe the new feature

Currently, it is unfeasible to enable Feature Metrics for large-scale applications. 
Whenever a metric change is triggered, this is what generally happens:

1. fetch serialized PHP object from the cache
1. deserialize the object
1. do some filtering
1. generally, increment some counter
1. serialize object
1. save it to the cache

If we have hundreds of users per second, it gets heavy on the CPU and the network to store metric information if, i.e., Redis is used. 
In a real-world application we tried, only the Metric I/O consumes 100Mb/s of network traffic to Redis. This is 10x more than the whole application itself.

To make it more resource-friendly:
* store plain values to avoid serialization
* make use of metric keys to avoid the need for PHP objects
* etc.

Unable to use the Metrics feature on more significant projects

_No response_

_No response_

RikudouSage:
# Description

Instead of serializing the whole object to cache, a simple (and small) string which can be reconstructed to the metrics bucket object is serialized instead. Should massively lower the size of the metrics in the cache storage.

Fixes #197 

## Type of change

# How Has This Been Tested?

- [x] Unit tests

# Checklist:

### DIFF
--- a/src/Metrics/DefaultMetricsHandler.php
+++ b/src/Metrics/DefaultMetricsHandler.php
@@ -38,6 +38,9 @@ final readonly class DefaultMetricsHandler implements MetricsHandler
         $bucket = null;
         if ($cache->has(CacheKey::METRICS_BUCKET)) {
             $bucket = $cache->get(CacheKey::METRICS_BUCKET);
+            if (is_string($bucket)) {
+                $bucket = $this->configuration->getMetricsBucketSerializer()->deserialize($bucket);
+            }
             assert($bucket instanceof MetricsBucket || $bucket === null);
         }
 
@@ -71,6 +74,6 @@ final readonly class DefaultMetricsHandler implements MetricsHandler
     private function store(MetricsBucket $bucket): void
     {
         $cache = $this->configuration->getMetricsCache();
-        $cache->set(CacheKey::METRICS_BUCKET, $bucket);
+        $cache->set(CacheKey::METRICS_BUCKET, $this->configuration->getMetricsBucketSerializer()->serialize($bucket));
     }
 }
--- a/shell.nix
+++ b/shell.nix
@@ -1,0 +1,35 @@
+{ pkgs ? import <nixpkgs> {} }:
+pkgs.mkShell {
+    nativeBuildInputs = with pkgs.buildPackages;
+    let
+        unstable = import (builtins.fetchTarball https://github.com/nixos/nixpkgs/tarball/master) {};
+        php83 = unstable.php83.buildEnv {
+            extensions = ({ enabled, all }: enabled ++ (with all; [
+                apcu
+                ctype
+                iconv
+                intl
+                mbstring
+                pdo
+                redis
+                xdebug
+                xsl
+            ]));
+            extraConfig = ''
+                memory_limit=8G
+                xdebug.mode=debug
+                apc.enable_cli=1
+                apc.enabled=1
+            '';
+        };
+     in
+     [
+        php83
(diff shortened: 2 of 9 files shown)


RikudouSage:
@vbartusevicius Does this work for you regarding #197?

RikudouSage:
As another alternative, consider using a different cache adapter for metrics, for example APCu - metrics are non-critical, so if something gets lost because of a restarting process, nothing really important should be lost (or you can implement a procedure where the metrics bucket gets flushed to server in case of shutdown). 

By using APCu you would cut the network overhead of Redis and it's not important to have a shared cache for metrics in particular. Alternatively take a look at all the [Symfony adapters](https://github.com/symfony/cache/tree/7.0/Adapter).

You can use something like this (if you use `symfony/cache`):

```php
<?php

use Symfony\Component\Cache\Adapter\ApcuAdapter;
use Symfony\Component\Cache\Psr16Cache;
use Unleash\Client\UnleashBuilder;

$unleash = (new UnleashBuilder())
    ->withMetricsCacheHandler(new Psr16Cache(new ApcuAdapter()))
    // your other builder options
    ->build()
;
```

src/Metrics/DefaultMetricsHandler.php
```diff
@@ -38,6 +38,9 @@ private function getOrCreateBucket(): MetricsBucket
         $bucket = null;
         if ($cache->has(CacheKey::METRICS_BUCKET)) {
             $bucket = $cache->get(CacheKey::METRICS_BUCKET);
+            if (is_string($bucket)) {
```
sighphyre, line 41: Quick double check - this should allow a user to replace an existing version connected to a cache that still has the old format and still work?

sighphyre (approved):
LGTM!

Would love to see the data on how much denser this is!

RikudouSage:
Some quick tests I did today:

=== Toggle amount: 1 ===
Old serialization size: 1,538 bytes
New serialization size: 31 bytes
Saving: 97.98%

=== Toggle amount: 10 ===
Old serialization size: 11,016 bytes
New serialization size: 121 bytes
Saving: 98.90%

=== Toggle amount: 100 ===
Old serialization size: 105,967 bytes
New serialization size: 1,111 bytes
Saving: 98.95%

=== Toggle amount: 1,000 ===
Old serialization size: 1,057,268 bytes
New serialization size: 11,911 bytes
Saving: 98.87%

=== Toggle amount: 10,000 ===
Old serialization size: 10,588,269 bytes
New serialization size: 128,911 bytes
Saving: 98.78%

=== Toggle amount: 100,000 ===
Old serialization size: 106,078,270 bytes
New serialization size: 1,388,911 bytes
Saving: 98.69%

src/Metrics/DefaultMetricsHandler.php
RikudouSage: Yes, that's exactly its purpose, before this it would return the object itself, now it returns the serialized string.

sighphyre: Fantastic!

vbartusevicius:
Thanks, @RikudouSage - these size savings are amazing. We are patiently waiting for the release, including this improvement!

RikudouSage:
@vbartusevicius It was already released a while ago, I just forgot to do the release on GitHub, but composer should have picked up the new version already.

PR #200, opened 2024-02-09
Chore: Update all actions to latest versions

RikudouSage:
# Description

Updates all actions to their latest versions to get rid of stuff like warnings that actions use outdated runtimes.


sighphyre (approved):
LGTM

PR #204, opened 2024-05-20
Chore: Modernize code using Rector

RikudouSage:
# Description

Modernizes the code using automated tools. Most commonly adds the PHP 8.3 #[Override] attribute. The attribute checks at compile time that the method is indeed overriding something and throws an error if it's not.

Other changes include:

- all global namespace classes are imported for consistency
- anonymous functions have been changed to short closures where applicable
- `get_class()` function call has been replaced with static `::class`
- unused caught exception variables have been removed
- the `Stringable` interface is now implemented explicitly instead of implicitly

Note that all of this will be automatically reversed when releasing for older versions.

# Checklist:


sighphyre (approved):
LGTM

PR #207, opened 2024-06-24
Chore: Fix new rule introduced by static analysis framework

RikudouSage:
# Description

The static analysis framework we use introduced a stricter check, this makes it conform to that.


sighphyre (approved):
LGTM

PR #209, opened 2024-06-25
Chore: Specify php version for releases

RikudouSage:
# Description

Specifies a PHP version for the release script, the release would currently fail because the default version on GitHub actions is apparently lower than 8.3.


sighphyre (approved):
LGTM

PR #211, opened 2024-07-01
Feat: Add possibility to refresh Cache, regardless of TTL

michael-schmetter:
# Description

This pull request introduces a new refreshCache functionality, designed to proactively refresh specific cached items before their TTL (time-to-live) expires. This enhancement ensures that our application maintains up-to-date data in the cache, minimizing the likelihood of cache misses.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [x] Unit tests

# Checklist:

### DIFF
--- a/src/Repository/DefaultUnleashRepository.php
+++ b/src/Repository/DefaultUnleashRepository.php
@@ -116,64 +116,19 @@ final readonly class DefaultUnleashRepository implements UnleashRepository
     {
         $features = $this->getCachedFeatures();
         if ($features === null) {
-            $data = null;
-            if (!$this->configuration->isFetchingEnabled()) {
-                if (!$rawData = $this->getBootstrappedResponse()) {
-                    throw new LogicException('Fetching of Unleash api is disabled but no bootstrap is provided');
-                }
-            } else {
-                $request = $this->requestFactory
-                    ->createRequest('GET', (string) Url::appendPath($this->configuration->getUrl(), 'client/features'))
-                    ->withHeader('UNLEASH-APPNAME', $this->configuration->getAppName())
-                    ->withHeader('UNLEASH-INSTANCEID', $this->configuration->getInstanceId())
-                    ->withHeader('Unleash-Client-Spec', '4.3.2')
-                ;
-
-                foreach ($this->configuration->getHeaders() as $name => $value) {
-                    $request = $request->withHeader($name, $value);
-                }
-
-                try {
-                    $response = $this->httpClient->sendRequest($request);
-                    if ($response->getStatusCode() === 200) {
-                        $rawData = (string) $response->getBody();
-                        $data = json_decode($rawData, true);
-                        if (($lastError = json_last_error()) !== JSON_ERROR_NONE) {
-                            throw new InvalidValueException(
-                                sprintf("JsonException: '%s'", json_last_error_msg()),
-                                $lastError
-                            );
-                        }
-                        $this->setLastValidState($rawData);
-                    } else {
-                        throw new HttpResponseException("Invalid status code: '{$response->getStatusCode()}'");
-                    }
-                } catch (Exception $exception) {
-                    $this->configuration->getEventDispatcher()->dispatch(
-                        new FetchingDataFailedEvent($exception),
-                        UnleashEvents::FETCHING_DATA_FAILED,
-                    );
-                    $rawData = $this->getLastValidState();
-                }
-                $rawData ??= $this->getBootstrappedResponse();
-                if ($rawData === null) {
-                    throw new HttpResponseException(sprintf(
-                        'Got invalid response code when getting features and no default bootstrap provided: %s',
-                        isset($response) ? $response->getStatusCode() : 'unknown response status code'
-                    ), 0, $exception ?? null);
-                }
-            }
-
-            if ($data === null) {
-                $data = json_decode($rawData, true);
-            }
-
-            assert(is_array($data));
-            $features = $this->parseFeatures($data);
-            $this->setCache($features);
+            $features = $this->fetchFeatures();
         }
 
         return $features;
+    }
+
+    /**
+     * @throws InvalidArgumentException
+     * @throws ClientExceptionInterface
+     */
+    public function refreshCache(): void
+    {
+        $this->fetchFeatures();
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/DefaultProxyUnleash.php
```diff
@@ -54,4 +55,13 @@ public function getVariant(string $featureName, ?Context $context = null, ?Varia
 
         return $variant;
     }
+
+    /**
+     * @codeCoverageIgnore
+     */
```
RikudouSage: This should be removed and the method should be called in [CoverageOnlyTest](https://github.com/Unleash/unleash-client-php/blob/main/tests/CoverageOnlyTest.php). The reasoning is that the code gets changed by an automated tool, so every line should at least be called to check that the tool did not introduce any fatal breaks.

src/Repository/DefaultUnleashRepository.php
```diff
@@ -538,4 +493,71 @@ private function parseVariants(array $variantsRaw): array
 
         return $variants;
     }
+
+    /**
+     * @throws ClientExceptionInterface
+     * @throws InvalidArgumentException
+     *
+     * @return void
```
RikudouSage: There are two returns, delete this one, please.

src/Repository/UnleashRepository.php
```diff
@@ -12,4 +12,6 @@ public function findFeature(string $featureName): ?Feature;
      * @return iterable<Feature>
      */
     public function getFeatures(): iterable;
+
+    public function refreshCache(): void;
```
RikudouSage: This would mean a breaking change, it needs to be added as a phpdoc block above the interface like this:

```php
/**
 * @method void refreshCache()
 */
```

Afterwards, when a new major version is released, it will be moved to a real method.

src/Unleash.php
```diff
@@ -14,4 +14,6 @@ public function isEnabled(string $featureName, ?Context $context = null, bool $d
     public function getVariant(string $featureName, ?Context $context = null, ?Variant $fallbackVariant = null): Variant;
 
     public function register(): bool;
+
+    public function refreshCache(): void;
```
RikudouSage: This IMO doesn't really belong on the Unleash interface, it should be called as part of the repository.

tests/Repository/DefaultUnleashRepositoryTest.php
```diff
@@ -168,6 +168,33 @@ public function testCacheWithNoFeatures()
...
+
+        sleep(3);
+        $repository->refreshCache();
+        sleep(3);
+
+        self::assertEquals(1, $this->mockHandler->count());
+
+        //Cache entry should have been invalidated by now (3+3 > TTL) if we didn't refresh it
+        $repository->getFeatures();
+        self::assertEquals(1, $this->mockHandler->count());
```
RikudouSage: Is the count here not cumulative? As in, weren't two http requests made during this test?

RikudouSage (changes requested):

tests/Repository/DefaultUnleashRepositoryTest.php
michael-schmetter: I also was a bit puzzled, but the `count()` method is returning the remaining items in the queue. You could argue to start with 2 response in the queue and check for 0 here...

```
    /**
     * Returns the number of remaining items in the queue.
     */
    public function count(): int
    {
        return \count($this->queue);
    }
```

src/Unleash.php
michael-schmetter: I don't like that either, do you have an idea how to access the functionality then? I only did that to expose it to users.

src/Repository/DefaultUnleashRepository.php
michael-schmetter: Fixed

src/Repository/UnleashRepository.php
michael-schmetter: Fixed

tests/Repository/DefaultUnleashRepositoryTest.php
RikudouSage: That would be better IMO, this got me all confused. You can also use `assertCount($expectedAmount,$this->requestHistory)`.

src/Unleash.php
RikudouSage: Currently you would need to provide your own repository to have access to that, so that's an option. Other than that, rework of the builder might be due so that it can return not only the `Unleash` object, but also the intermediate objects.

@sighphyre What's your opinion?

src/DefaultProxyUnleash.php
michael-schmetter: Will do that, when we decide how to progress with the the Unleash interface

src/Unleash.php
sighphyre: > This IMO doesn't really belong on the Unleash interface, it should be called as part of the repository.

Very much agree. I don't think it belongs in the top level API, and I think the `BadMethodCallException` in the `DefaultProxyUnleash` is a warning sign that the abstraction is subtly off. 

I can't really see forced cache updates being called from the same abstraction layers as `isEnabled` /`getVariant` in user code (entirely open to the idea that I just don't understand how PHP apps are architected though). Forced cache updates feels like a thing I want called in another thread/process, to me that says that right now the `Repository` is the correct layer for that to live and that maybe this PR is currently a stepping stone to having an updater/scheduler component

tests/Repository/DefaultUnleashRepositoryTest.php
michael-schmetter: Changed

src/DefaultProxyUnleash.php
```diff
@@ -54,4 +55,13 @@ public function getVariant(string $featureName, ?Context $context = null, ?Varia
...
+    #[Override]
+    public function refreshCache(): void
+    {
+        throw new BadMethodCallException('This method is not supported in the proxy implementation');
```
sighphyre: Ideally I'd like to see no changes to this file. Having a method that breaks at runtime doesn't feel good to me, I think we can move this method down into the Repository perhaps, at least that way, it's clear to end users that this isn't something you can do with the Proxy instance

src/Unleash.php
RikudouSage: @michael-schmetter Please remove the method from the `Unleash` interface altogether, the PR is fine without it, solving how to get the Repository object out of the builder should be a separate PR.

Currently the only way to have access to repository is by manually constructing the whole object (as seen [in the example](https://github.com/Unleash/unleash-client-php/blob/main/examples/07-manual-unleash-object.php)).

In a separate PR we can do one of two things:

- make it possible to set a custom repository on the builder
- rework the builder to return intermediate objects as well

If you'd like to tackle it, let me know, otherwise I can do the PR myself.

src/DefaultProxyUnleash.php
RikudouSage: Resolving this due to the other conversation where we decided to go in a different direction.

src/DefaultProxyUnleash.php
michael-schmetter: Removed it

src/DefaultProxyUnleash.php
michael-schmetter: Removed the changes to the Unlesh interface

src/Unleash.php
michael-schmetter: @RikudouSage 
I updated the PR and removed all superfluous changes that were related to the `Unleash` interface.

I can do a PR, but let's quickly connect, what you would have envisioned.

something like 

```
list($unleash, $repository) = UnleashBuilder::create()
    ->withInstanceId('Some instance id')
    ->buildWithRepository();
```

Or would you go another route?

RikudouSage: @michael-schmetter I'd go in the direction of there being multiple methods, like `buildRepository()` and `build()` and possibly others. The `build()` would internally reuse the `buildRepository()` etc.

Edit: I meant on the UnleashBuilder, just to be clear.

RikudouSage (approved):

PR #218, opened 2024-07-23
Fix: race condition in caching layer

sighphyre:
# Description

This fixes a bit of a subtle bug in the caching layer where the SDK can start to return empty sets of flags if the cache is invalidated between checking the cache status and doing the actual retrieve. This forces the read from cache to be atomic.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

Custom script - spin up 50 processes each running their own SDK, backed by Redis all checking a feature that should always be enabled. This typically produces a false result in 10-20 seconds. Apply fix rerun script for 20 minutes.

# Checklist:

### DIFF
--- a/src/Repository/DefaultUnleashRepository.php
+++ b/src/Repository/DefaultUnleashRepository.php
@@ -140,11 +140,10 @@ final readonly class DefaultUnleashRepository implements UnleashRepository
     {
         $cache = $this->configuration->getCache();
 
-        if (!$cache->has(CacheKey::FEATURES)) {
+        $result = $cache->get(CacheKey::FEATURES);
+        if ($result === null) {
             return null;
         }
-
-        $result = $cache->get(CacheKey::FEATURES, []);
         assert(is_array($result));
 
         return $result;


src/Repository/DefaultUnleashRepository.php
```diff
@@ -140,11 +140,10 @@ private function getCachedFeatures(): ?array
     {
         $cache = $this->configuration->getCache();
 
-        if (!$cache->has(CacheKey::FEATURES)) {
+        $result = $cache->get(CacheKey::FEATURES);
+        if ($result === null) {
```
sachinmaharjan, line 144: Could this $results from cache could have empty arrays like []?

RikudouSage (approved):
Looks good!

src/Repository/DefaultUnleashRepository.php
sighphyre: In the previous iteration yeah, in this, that shouldn't happen unless something puts an empty array in there. Assuming no bugs are present in the code, that should only ever happen if the SDK is using a token that's connected to a project that has no features

PR #221, opened 2024-07-25
Chore: Add an example for http timeout

### LINKED ISSUE #220: Feat: method `->withTimeout(secs)`

### Describe the new feature

Sometimes our self-hosted gitlab is unreachable due to some maintenance activities. In that case, it takes more than 1 min to return the cached value or the default value. Can we add method `->withTimeout(secs)`?

Sometimes our self-hosted gitlab is unreachable due to some maintenance activities. In that case, it takes more than 1 min to return the cached value or the default value.

set_time_limit

_No response_

RikudouSage:
# Description

Adds an example on how to add a http client timeout for the two most popular http clients.

Fixes #220 

# Checklist:

### DIFF
--- a/examples/10-timeout.php
+++ b/examples/10-timeout.php
@@ -1,0 +1,30 @@
+<?php
+
+use GuzzleHttp\Client;
+use GuzzleHttp\RequestOptions;
+use Symfony\Component\HttpClient\HttpClient;
+use Symfony\Component\HttpClient\Psr18Client;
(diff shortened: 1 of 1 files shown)


sighphyre (approved):
Sure looks good!

Small suggestion but since this is already importing `common`, might as add   `->withHeader('Authorization', $apiKey)` and do an `isEnabled("myFeature");` call

RikudouSage:
Added it, thanks.

PR #223, opened 2024-07-26
Chore/bump client spec version

sighphyre:
Moves the version of the client specification that's advertised to Unleash to 5.0.2 in order to match the version that this SDK is actually tested against.

@RikudouSage this is probably a cheap patch to get the versions matching again but ideally the version would always match and be enforced at build time. Most of the other SDKs now just clone a shallow snapshot of the repo to the tag that matches the source code when it's being run in CI. Are you deeply attached to using a submodule for the client spec because forcing the versions to match isn't quite as easy with the submodule

### DIFF
--- a/src/Unleash.php
+++ b/src/Unleash.php
@@ -9,7 +9,7 @@ interface Unleash
 {
     public const string SDK_VERSION = '2.5.1';
 
-    public const string SPECIFICATION_VERSION = '4.3.2';
+    public const string SPECIFICATION_VERSION = '5.0.2';
 
     public function isEnabled(string $featureName, ?Context $context = null, bool $default = false): bool;
 


nunogois (approved):

RikudouSage:
~~I wasn't aware that part needs updating, I think I didn't have any part in developing this.~~ I see this does not go into main, but a feature branch.

I'm not deeply attached to it, though I rather like it being linked. Shouldn't it just be possible to check what revision is the cloned repository at and if it does not match, fail? Should be just a few git commands, shouldn't it?

sighphyre:
> Shouldn't it just be possible to check what revision is the cloned repository at and if it does not match, fail? Should be just a few git commands, shouldn't it?

Yeah actually that sounds pretty reasonable and less invasive

PR #240, opened 2025-07-01
Feat: Make repository replaceable in the builder

### LINKED ISSUE #239: Feat: granular cache strategy

- [x] I have searched the existing issues

Hello there ! 👋

First of all, thank you for this amazing open-source service. The quality is outstanding, and it has truly transformed the way we work since we adopted it! 🚀

### Context
We've been using Unleash for several months now in a Laravel 11 project, with Redis as our cache driver. Along the way, we've encountered a few issues—some we've managed to solve ourselves, while others are a bit more complex (and would likely require modifying the vendor package).

### Past issues
Here are two caching strategies we've implemented to improve performance and stability:

We avoid making multiple calls to the Unleash service for the same feature flag. Instead, we cache the result in memory whenever possible, to reduce read pressure on Redis.

We no longer allow feature flag cache entries to expire. Expirations were causing "cache avalanches" when frequently accessed endpoints triggered simultaneous refreshes. Our current approach: a CRON job refreshes the cache every 3 minutes, and we set a higher TTL (which is now mostly irrelevant). This helps us avoid concurrency issues that were heavily impacting Redis.

### Our facing issue
Our current challenge lies in how the cache is structured. From what we’ve observed, all feature flags are stored under a single Redis key. As a result, even when checking just one feature flag, the system retrieves the entire set from Redis. This caching strategy becomes problematic especially when checking feature flags on high-traffic endpoints, leading to a super high reading operation on our Redis.

### Solution lead
Our (naive) idea is to rewrite the caching logic to store each feature flag under its own Redis key. This would make reads more efficient and reduce unnecessary data transfers.

Do you think this would be a viable approach? We’d love to hear your thoughts 🙏
Thanks again for the great work!

RikudouSage:
# Description

Makes it possible to do two things:

1. Entirely replace the repository with a custom solution
2. Easily create a decorator that wraps the built-in repository

Fixes #239 

## Type of change

# How Has This Been Tested?

- [ ] Unit tests

# Checklist:

### DIFF
--- a/examples/11-custom-repository.php
+++ b/examples/11-custom-repository.php
@@ -1,0 +1,64 @@
+<?php
+
+use Psr\Cache\CacheItemPoolInterface;
+use Symfony\Component\Cache\Adapter\FilesystemAdapter;
+use Unleash\Client\DTO\Feature;
+use Unleash\Client\Repository\UnleashRepository;
+use Unleash\Client\UnleashBuilder;
+
+require __DIR__ . '/_common.php';
+
+// first create a builder to later get the default repository
+$unleashBuilder = UnleashBuilder::create()
+    ->withAppName($appName)
+    ->withAppUrl($appUrl)
+    ->withInstanceId($instanceId)
+    ->withHeader('Authorization', $apiKey)
+;
+$originalRepository = $unleashBuilder->buildRepository();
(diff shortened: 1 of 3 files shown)


sighphyre (approved):
LGTM

PR #254, opened 2026-07-17
feat: add SDK flavour config

sighphyre:
Useful for wrapper SDKs to identify themselves, so we can track adoption in the wild


irstavr (approved):
I don't really detect any issues here - but again ..my php is ancient-knowledge :p
I trust you here 👍🏽  You're good to go! 🎉

PR #256, opened 2026-07-28
fix: rename sdkFlavour and sdkFlavourVersion

sighphyre:
Headers in Unleash are spelled with American English :(

This should keep the public API intact and still fix the bug


irstavr (approved):
the power of the dollar 🇺🇸

PR #260, opened 2026-08-19
chore: remove semver metadata from comparison

daveleek:
This removes metadata from semver versions both on context value and constraint value


sighphyre (commented):
The tests here are a bit too thin for my tastes. A bunch of the constraint stuff can get away with thin tests because they're covered by the client spec but this isn't taking on a new version of the client spec. 

So either we need to port those same tests here or we need to depend on a version or subset of that client spec that does the covering

sighphyre (approved):

PR #261, opened 2026-08-19
feat: add support for semver gte and lte constraints

daveleek:
Adds support for GTE and LTE constraint operators.


sighphyre (approved):
Cheers

